PR #36, opened 2015-09-11
Integration weights

fietew:
Add integration weights for regular secondary source geometries based on the distance between adjacent sources along the continuous boundary. Furthermore, the tapering window is normalized such that each source is weighted (on average) with 1.0 . Both modifications should prevent amplitude fluctuations between the reproduced soundfield for different primary source positions/directions.


hagenw:
Hi Fiete.
Nice idea to normalize the window function. This will give slightly different results depending on the array geometry, but should be better than nothing in all cases.

fietew:
I guess, I have to tackle this issue again: For virtual plane waves and planar/linear secondary source distributions normalizing the tapering window to one might be beneficial. However, for virtual point sources it leads to an undesired amplitude offset (originally the normalization was introduced to prevent this). I will have a look at it, again.

PR #39, opened 2015-09-22
Update dummy_irs()

hagenw:
In order to improve memory consumption and speeding up calculation I reduced the number of stored dirac pulses returned by `dummy_irs()` to only 4. It principle it should also work by using only 1, but this would need some changes in the other functions.
We could still think about if this would be worth the effort.
I tested the new version at the [example from the README](https://github.com/sfstoolbox/sfs#make-binaural-simulations-of-your-systems) and the difference between the results were <10^-18.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,15 +22,15 @@ support](https://github.com/sfstoolbox/sfs/releases/tag/1.2.0).
 
 ### Table of Contents
 
-**[Installation](#installation)**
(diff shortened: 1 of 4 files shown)


hagenw:
OK, I was now able to use only one Dirac impulse. I changed `findnearestneighbour()` in a way that it will return now only as many neighbors as available.

Would be nice if you could test it, too.

PR #43, opened 2015-10-06
establish line sink as default driving function for 2D wfs

fietew:
Having a focused point source as the default driving function in the 2-dimensional case is imho not very intuitive, since the line sources are used as secondary source. Therefore, I would suggest to use the model of the line sink as the default driving function.


hagenw:
This problem emerges from some general naming inconsistency. In all other cases we have always the same driving function for 2D and 3D as there should be no difference between the mathematical solutions. The problem with focused sources is that we have only the name "focused source" instead of "focused line source" and "focused point source". I think this is ok, as it is the way it is handled in theory. Nonetheless, I wanted to avoid to have a distinction between 2D and 3D inside the driving function and that is the reason why the default driving function is always a focused point source.

I agree, that this is not very intuitive and we should change it. The problem with your current implementation is that the code for the line source is repeated for the 2D and 3D case. It is not very nice, but maybe it is ok, and we simply live with it?

fietew:
The same driving functions for 2D and 3D is maybe concept, which we shall reassessed. Having e.g. a driving function for a virtual point source in a pure 2D scenario, where we have secondary line sources, does not make much sense imho: It is impossible in my view to create the amplitude decay 1/r of a point source with line sources having a decay of 1/sqrt(r).
I agree that the repetitions of the code is not very nice. One option could be to change the order of the if-clauses (first the driving functions, then the dimensionality). Second option may be to combine dimensionality and driving functions into one string and use a ``switch`` directive, which is able to handle multiple strings in one ``case``

hagenw:
Yes, it is correct, it is impossible to create a point source in a 2D setup and it is also impossible in most of the 3D setups to create a line source. The idea behind having those driving functions in the toolbox is that people can see themselves.

hagenw:
I found a way to incorporate this a little bit easier.

Now, we have this only for monochromatic sound fields. Could you implement it also for the time domain?

hagenw:
I have added no also the time domain version of the line sink focused source for 2D.
@fietew: could you please test.

fietew:
Where does this formula come from? see http://sfstoolbox.org/#equation-d.wfs.fs.ls

hagenw:
I guess it comes from http://sfstoolbox.org/en/latest/#equation-d.wfs.ls and should be its time inversed version. But we should remove `v` and replace it with `x0-xs` as for the D case. Do you agree?

hagenw:
I have corrected the documentation and going to merge this now. If there are still some errors, please create an issue and we fix it in the next release.

PR #44, opened 2015-10-09
add model for 3D dipole source

fietew:
As I am using it for some simulations, maybe this might be useful.


hagenw:
Could you add the same for the time domain, to be more consistent, please. And also add the equation to the PDF documentation document at https://github.com/sfstoolbox/sfs-documentation.

fietew:
Where exactly in the documentation should I add this? 2.4. Model-Based Rendering ?

hagenw:
Exactly.

PR #47, opened 2015-10-30
Make config struct mandatory

hagenw:
As our normal workflow is to create first a conf struct and disable the `SFS_config()` function by using `SFS_config_example()` I have removed the automatically calling of `SFS_config()` completely.

I only have to test if everything is working, otherwise the changes should be ready.


hagenw:
I have tested the new code and fixed a typo regarding the new config behavior, but also one general thing with replacing `range()` by `max()-min()`.

PR #49, opened 2016-01-11
Change Ambisonics order for even loudspeaker numbers

hagenw:
Before it was N/2, which is not identical to the cited literature (e.g. Ahrens 2012, p. 132). There it is N/2-1, which I changed the code accordingly.
Before applying this changes: was this just a mistake, or did we choose the wrong version on purpose?


hagenw:
This was a bug as the modal spectrum will still overlap with N/2.

PR #51, opened 2016-01-12
fix recursive call of selection function and remove time consuming vector product

fietew:
For some applications, e.g. local WFS, the secondary source selection is very slow. This is due to the recursive calls of ``secondary_source_selection`` and the usage of the - imho not necessary - function ``vector_product``. Furthermore ``xs`` does not need to be replicated to perform ``x0-xs``.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -6,6 +6,7 @@ Changes in the Sound Field Synthesis-Toolbox. Recent changes on top.
     - add new start message
     - fix handling of 0 in least squares fractional delays
     - fix NFC-HOA order for even loudspeaker numbers to N/2-1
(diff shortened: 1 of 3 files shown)


hagenw:
Cool, I was not aware that it is possible to achieve it with something like nk(:).

I have tested it with running
```Matlab
>> test_driving_functions(1)
```
and the output seems to be fine. It would be nice if we could have a better way of testing, if such changes gave exactly the correct results.
Do you have an idea, how to write a good test routine for this?

fietew:
We could create a new testing function which checks the secondary source selection criterion by hand.

Issue #55, opened 2016-01-25
Propagation direction of NFC-HOA in the time domain


narahahn:
In [driving_function_imp_nfchoa.m](https://github.com/sfstoolbox/sfs/blob/master/SFS_time_domain/driving_function_imp_nfchoa.m#L124), there is a if-request to solve the problem of wrong propagation direction. I found that this comes from the error in computing the input for spatial IFFT ([here](https://github.com/sfstoolbox/sfs/blob/master/SFS_time_domain/driving_function_imp_nfchoa.m#L124)). According to the Eq. (5) and (6) of [Spors et al, 2011](http://www.int.uni-rostock.de/fileadmin/user_upload/publications/spors/2011/Spors_etal_WASPAA_2011_efficient_NFCHOA.pdf) the exponential term should have negative exponent.
Should I handle this issue myself?

PR #58, opened 2016-01-25
Correct sign in NFC-HOA time domain driving function

hagenw:
I changed the NFC-HOA time domain driving function, according to #55. But there are still some problems.

Try the following commands:

```Matlab
conf = SFS_config;
conf.nfchoa.order = 32;
sound_field_imp_nfchoa([-2 2],[-2 2],0,[1 0 0],'pw',300,conf)
sound_field_imp_nfchoa([-2 2],[-2 2],0,[2.5 0 0],'ps',300,conf)
```

Here are the results from before the change:

nfchoa_pw_before
nfchoa_ps_after

And here after the changes:

nfchoa_pw_after

nfchoa_ps_before

As you can see, the results for the point source is still correct, but in the case of the plane wave we have no a wrong direction.

### DIFF
--- a/SFS_time_domain/driving_function_imp_nfchoa.m
+++ b/SFS_time_domain/driving_function_imp_nfchoa.m
@@ -80,14 +80,7 @@ order = nfchoa_order(nls,conf);
 
 % Correct position of source for off-center arrays
 xs(1:3) = xs(1:3)-X0;
-% If-request as a workaround for the right direction of the sound field
-if strcmpi(src,'pw')
-    [theta_src, r_src] = cart2pol(-xs(1),xs(2));
-elseif strcmpi(src,'ps')
-    [theta_src, r_src] = cart2pol(xs(1),-xs(2));
-else
-    [theta_src, r_src] = cart2pol(xs(1),xs(2));
-end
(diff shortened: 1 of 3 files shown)


narahahn:
I just found another error in the plane wave driving function. The `(-1)^abs(m)` term of Eq. (10) is missing ([Spors et al, 2011](http://www.int.uni-rostock.de/fileadmin/user_upload/publications/spors/2011/Spors_etal_WASPAA_2011_efficient_NFCHOA.pdf)). But you have to be careful, when fixing this. If you just multiply  `sos` with `(-1)^abs(m)` the numerator and denominator of the second-order section are multiplied simultaneously, and cancelled out. One possibility to do it correctly would be:

```
sos(1,1:3) = sos(1,1:3) * (-1)^abs(order);
```

hagenw:
OK, I [implemented it](https://github.com/sfstoolbox/sfs/blob/nfchoa_propagation_direction/SFS_time_domain/driving_functions_imp/driving_function_imp_nfchoa_pw.m#L101) and it seems to work now. Could you also test it, please.

narahahn:
Great. Also works in my Matlab.

PR #62, opened 2016-02-05
Return wfs delay

hagenw:
This is a follow up on #59.

The main goal was fomulated by @fietew as:

> Imho, it would be easier to return the pre-delay which has been added to the driving function by a particular processing step, e.g. fractional delay, wfs prefilter, focused source, plane wave, etc..
Furthermore, I would prefer that t=0 corresponds to the t=0 of the virtual source, i.e. where the sound "starts" to radiate from a point/focused source or when the plane wave passed its point of zero phase.

In order to achieve this I did the following things:

1. Add an additional `prefilter_delay` return parameter to `wfs_preequalization()`, which returns the delay that is added by the applied pre-filter.
2. Add an additional `delay_offset` return parameter to `driving_function_imp_wfs()`. In the case of `conf.removedelay=1` it returns `prefilter_delay`. In the case of `conf.wfs.removedelay=0` it returns `secondary_source_diameter(conf)/c + prefilter_delay`
3. I used this directly in `sound_field_imp_wfs()`:
```Matlab
[d,~,~,delay_offset] = driving_function_imp_wfs(x0,xs,src,conf);
t = t + delay_offset*fs;
```

**Result**
That means if you use `sound_field_imp_wfs()` with `conf.wfs.removedelay=1` you will have the first activity at the secondary sources for `t = 0`. With `conf.wfs.removedelay=0` you will have the first activity of the virtual sources for `t = 0`.

**Example**
```Matlab
conf = SFS_config;
conf.wfs.removedelay = 1;
conf.plot.usedb = 1;
sound_field_imp_wfs([-2 2],[-2 2],0,[0 0.5 0 0 -1 0],'fs',0,conf)
```

wfs_fs_removedelay1

```Matlab
conf = SFS_config;
conf.wfs.removedelay = 0;
conf.plot.usedb = 1;
sound_field_imp_wfs([-2 2],[-2 2],0,[0 0.5 0 0 -1 0],'fs',0,conf)
```

wfs_fs_removedelay0

**Open questions**

- @fietew : as you can see I have only added the return value to the driving function, not to the delayline etc. Could you maybe add it to the other functions?
- @trettberg : I have added a [return value also for the IIR filter](https://github.com/sfstoolbox/sfs/blob/return_wfs_delay/SFS_time_domain/wfs_preequalization.m#L88) with `delay = conf.wfs.hpreIIRorder/2 / fs;` I guess this is not correct, could you please correct it.

### DIFF
--- a/SFS_time_domain/driving_function_imp_wfs.m
+++ b/SFS_time_domain/driving_function_imp_wfs.m
@@ -1,24 +1,26 @@
-function [d,delay,weight] = driving_function_imp_wfs(x0,xs,src,conf)
+function [d,delay,weight,delay_offset] = driving_function_imp_wfs(x0,xs,src,conf)
 %DRIVING_FUNCTION_IMP_WFS_25D calculates the WFS weighting and delaying
 %
 %   Usage: [d,delay,weight] = driving_function_imp_wfs(x0,xs,src,conf)
 %
 %   Input parameters:
-%       x0      - positions and directions of secondary sources / m [nx6]
-%       xs      - position of virtual source or direction of plane wave / m
-%                 [1x3]
-%       src     - source type of the virtual source
-%                     'pw' - plane wave (xs, ys are the direction of the
-%                            plane wave in this case)
-%                     'ps' - point source
-%                     'ls' - line source
-%                     'fs' - focused source
-%       conf    - configuration struct (see SFS_config)
+%       x0              - positions and directions of secondary
+%                         sources / m [nx6]
+%       xs              - position of virtual source or direction of
+%                         plane wave / m [1x3]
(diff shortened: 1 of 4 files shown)


trettberg:
> I have added a return value also for the IIR filter
I think it's best to return a delay of zero, instead of something like maximum group delay.

btw: The IIR case in `wfs_preequalization` had some issues. It should hopefully work now.

fietew:
As the new version of the ``delayline`` is still under construction (see #50) I would prefer to add the ``delay_offset`` return value to this function after the merge of the new delayline.

hagenw:
OK, I have created #64 in order to not forget it and will merge this pull request in the meantime.

PR #63, opened 2016-02-10
Representation of sound fields and driving functions as circular and spherical harmonics

fietew:
Features:
* decompose point sources, line sources and plane waves into circular and/or spherical basis functions
* driving functions for WFS and NFC-HOA to reproduce sound fields expressed by spherical and/or circular basis functions
* spherical/circular harmonics transforms of driving functions

TODOs:
* convert testing functions into validation functions
* switch to recently introduced mandatory conf struct
* adapt sound field functions to support recently introduced non-regular grids


fietew:
Maybe the following function could be replace by their recurrence relation
``sphbesselh_derived``
``sphbesselj_derived``
``sphbessely_derived``
``besselh_derived``
``besselj_derived``
``bessely_derived``

Also this might not be necessary:
``isargsquaredinteger.m``

hagenw:
Can you please make sure that you use always 4 spaces as tab in for loops etc. and also that you have no spaces between `,` and arguments in function calls:
`f(a, b, c)` => `f(a,b,c)`.
This would be nice to maintain consistency across all functions in the Toolbox.

fietew:
I added it to the TODO list

hagenw:
If I remember correctly, we included some (or most?) of your proposals by other pull requests.
What is the current status and plan for this pull request?

fietew:
I would suggest to wait for the upcoming discussions on the DAGA poster for SFS Toolbox. Then I will have a more holistic view on what is already implemented and what might be taken from this PR.

PR #73, opened 2016-03-02
Fix 3D-WFS time domain driving function for virtual line source

trettberg:
This uses the correct distance from virtual line source to secondary sources.
(We fixed the mono-frequent case some weeks ago. This is just the same for the impulsive case.)


hagenw:
You are right, I forgot to complain about the missing corresponding changes in time domain.
Thanks for implementing this.

PR #74, opened 2016-03-08
Change removedelay wording

hagenw:
As we not really remove the delay completely in our `conf.wfs.removedelay` settings, I propose to change this setting to

``` Matlab
% === Time Domain Implementation ===
% Adjust the starting time in WFS-time domain driving functions.
% This can be set to
%   'system'   - the first secondary source will be active at t=1
%   'source'   - the virtual source will be active at t=1
% Setting it to 'system' is most convenient when simulating single sources as
% you will always see activity in the sound field for t>0. Setting it to
% 'source' helps you to simulate different sources as you can time align them
% easily. Note, that for virtual sources outside of the array this can mean you
% will see no activity inside the listening area until the time has passed, that
% the virtual source needs from its position until the nearest secondary source.
conf.wfs.t0 = 'system'; % string
```


fietew:
In the documentation: why ``t=1`` instead of ``t=0``? Is it a typo?

hagenw:
At t=0 you the extension of the activity is more or less infinite small, so t=1 should be the first sample where you can see activity. But if it is confusing we could also change it to t=0.

trettberg:
Changed to `t=0`. At least for the `source` case, this is definitely more accurate. Also added a sentence of warning that placement of focussed source may be restricted in this setting.

hagenw:
Cool, thanks, in my opinion this could then be merged, or do you have a better proposal for the used names?

trettberg:
I think it's fine this way.
(At some point in time, we/I should maybe add an example for documentation purpose. But that will have to wait.)

hagenw:
I added already a slightly modified version of your example to the time-domain section, see:
 https://github.com/sfstoolbox/sfs#simulate-time-snapshots-of-sound-fields (and scroll down)
NOTE: the README is for the master and includes the removedelay terminology. I also changed this in this pull request.

trettberg:
Oh, cool! Then I'd think it's ready to merge.

PR #75, opened 2016-03-11
retarded 3D Green's function uniformly moving point source

fietew:
to test:
`sound_field_mono([-2,2],[-2,2],0,[0,1,0,100,0,0,1],'mps',1,1000,conf)`

TODO:
for the moment, i had to deactivate the check for the secondary sources' orientation vector to be normalized, as the velocity of the moving source is encoded in the magnitude of the sound source. Any ideas regarding a good solution?


hagenw:
I see at least two easy solutions:

1. Use a `conf` entry for the velocity
2. Disable the checking of `nx0` and ensure that `nx0` is really a direction vector at those places where it has to be one (for example by using `direction_vector(nx0)`)

Any preference for one of the two solutions?

fietew:
2nd solution seems the better one as we should avoid additional entries in the `conf` struct if possible.

hagenw:
Yes, I'm also in favor of the 2nd solution. The only downside is that it will probably affect a lot of files.

Do you also plan to add a time-domain version of the moving source, or will this be to complicated?

fietew:
The time-domain version would require time-variant convolution and time-variant driving signals. This would require a complete new component in the toolbox. So far, I have no idea, how this could be included in the current framework.

hagenw:
Should we include then only the monochromatic version. I don't see that we develop the time domain version at the moment.

fietew:
So we remove the check for the direction of the secondary sources?

PR #77, opened 2016-04-19
add modal weighting to time-domain nfchoa

fietew:
in order to study the effect of different modal windows on the reproduced sound field

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -2,6 +2,7 @@ Changes in the Sound Field Synthesis-Toolbox. Recent changes on top.
 
 2.2.0 ()
     - fix impulse response interpolation for three points
+    - add the ability to apply modal weighting window to NFC-HOA in time domain
 
 2.1.0 (10. March 2016)
 
--- a/SFS_config.m
+++ b/SFS_config.m
@@ -259,10 +259,10 @@ conf.nfchoa.wtype = 'rect';  % string
 %   'kaiser', 'kaiser-bessel'  - Kaiser aka. Kaiser-Bessel window
 conf.nfchoa.wparameter = 0.0;  % float
 % scalar parameter for window, if applicable. Effect for distinct window:
-%   'rect'    - no effect 
-%   'kaiser'  - trade of between main-lobe width and side-lobe levels. 0.0
-%               results in the rectangular window and the smallest main-lobe
-%               width.
+%   'rect'    - no effect
+%   'kaiser'  - [0,inf]. trade-off between main-lobe width and side-lobe levels.
+%               0.0 results in the rectangular window and the smallest main-lobe
+%               width. infinity results in a dirac impulse.
 
 
 %% ===== Local Sound Field Synthesis =====================================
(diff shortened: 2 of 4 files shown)


hagenw:
I have corrected a small error and changed a little bit the behavior of the default parameters to be in line with the rest of the Toolbox, hopes this is ok with you.

At the moment the `modal_weighting()` function has only `Input parameters` and `Output parameters` in the comments at the beginning. Have you a good idea what we could add for a more detailed explanation, that maybe could roughly explain what it is doing to someone not familiar with the concept of modal weighting?

For adding a reference I guess we have to wait until you publish some results?

fietew:
Maybe we should use the `varargin` functionality of MATLAB instead of assigning `ndft` to `conf`. I think it's quite hard to understand, what happens, if only two arguments are given.

I could add some reference for the kaiser window function.

hagenw:
I have updated the parameter handling a little bit, but did not change the behavior to `varargin` as the current behavior is the standard procedure done in SFS, see [this example](https://github.com/sfstoolbox/sfs/blob/master/SFS_plotting/plot_sound_field.m#L60).

There is a [TODO left](https://github.com/sfstoolbox/sfs/blob/nfchoa_weighting/SFS_general/modal_weighting.m#L85), could you fix this and add the reference you mentioned, this would be nice.

hagenw:
I added a [small description](https://github.com/sfstoolbox/sfs/blob/nfchoa_weighting/SFS_general/modal_weighting.m#L17), maybe you can check if it is correct.

Then, I was wondering about the usage of the `conf.nfchoa.parameter` entry. If I see it correctly, a setting of `conf.nfchoa.wtype = 'kaiser'` and `conf.nfchoa.parameter = 0.0` will be the same as using `conf.nfchoa.wtpye = 'rect'`.
Are you planing to integrate more windows? Otherwise I would propose to use only the Kaiser-Bessel window and provide just one conf parameter, something like `conf.nfchoa.modal_window` which can go from 0 to 1.

fietew:
Yes, I would like to add some more windows in the future. ``conf.nfchoa.parameter`` is not that intuitive, but the many window type have one parameter, but the meaning of the parameter is quite different. Furthermore, the parameter of the Kaiser window can go from 0.0 to infinity. I will add that to the comments of the ``SFS_config``

hagenw:
OK, then I agree to have the setting in `conf`.
I'm still not completely happy with the naming, what do you think of the following proposal?
`conf.nfchoa.wtype` => `conf.nfchoa.modal_window`
`conf.nfchoa.wparameter` => `conf.modal_window_strengths` or `conf.modal_window_parameter`

fietew:
I think the window parameter should still have ``conf.nfchoa.`` in front. ``conf.modal_window_strengths`` is misleading.
If I had to chose, I would use ``conf.nfchoa.modal_window`` and ``conf.nfchoa.modal_window_parameter``

hagenw:
Sorry, my fault, I meant `conf.nfchoa.modal_window_parameter`.
I have changed the code accordingly, could you test if it is still working.

Currently the `modal_weighting` is only applied in the [temporal NFC-HOA driving function](https://github.com/sfstoolbox/sfs/blob/nfchoa_weighting/SFS_time_domain/driving_function_imp_nfchoa.m#L127). Is it also possible to apply it to the mono-frequent driving function or do we have to change the NFC-HOA structure (as you proposed) before?

fietew:
Yes, I had a look at the current monochromatic driving functions and it would be quite complicated to add the weightings there. With a new structure, i.e. 

1. compute the circular/spherical basic expansion of the sound field,
2. compute the circular/spherical harmonics coefficients of the driving functions,
3. convert it to the loudspeakers driving function, 

this would be easier. The modal weights could be applied directly in the second step.

fietew:
I tested it with a few parameter combination and everything looked fine for me.

PR #78, opened 2016-04-26
Add automatic scaling to loudspeaker weights for plotting

hagenw:
With 7b46fff0839585b595f88a95ffc84e5fb3c2f9b8￼ we introduced a new weighting to the loudspeakers, which results in relatively low values (<<1) for standard setups. This is of course not a problem, but has a small effect on the plotting of real loudspeaker symbols as those are colored accordingly to the absolute weight of the loudspeaker.

Here is an example from the current master:
```Matlab
conf = SFS_config;
conf.plot.realloudspeakers = 1;
sound_field_mono_wfs([-2 2],[-2 2],0,[0 2.5 0],'ps',800,conf)
```

ls_w_old

As you can see more or less all loudspeakers are white.

In this pull request I added an automatic scaling of the loudspeakers (that the maximum weight to 1) to the `draw_loudspeaker` function, which results in this plot by executing the same commands as above.

ls_w_new

I'm not sure yet if this is the best solution for more complex cases, but I think it is better than before.

### DIFF
--- a/SFS_plotting/draw_loudspeakers.m
+++ b/SFS_plotting/draw_loudspeakers.m
@@ -13,7 +13,8 @@ function draw_loudspeakers(x0,dimensions,conf)
 %   DRAW_LOUDSPEAKERS(x0,dimensions,conf) draws loudspeaker symbols or filled
 %   points at the given secondary source positions. This can be controlled by
 %   the conf.plot.realloudspeakers setting. The loudspeaker symbols are pointing
-%   in their given direction.
+%   in their given direction and are colored accordingly to their weights,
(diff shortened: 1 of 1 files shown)


fietew:
I think its a good idea. In te context of WFS and the secondary source selection its also nice, as the active secondary source would be dark/black and the inactive are white.

Maybe it's a good idea to use the maximum of the absolute value instead of the maximum. I don't know, why there should be any negative loudspeakers weights at the moment. Maybe someone decides to implement some ambisonics panning functions in the future, which may have negative weights.

hagenw:
OK, I changed it and will merge it into the master.

PR #80, opened 2016-05-11
Change license to MIT

hagenw:
I updated the LICENSE using [this script](https://github.com/sfstoolbox/maintenance-tools/blob/master/change_copyright_to_mit.sh).

Could you check, if you agree with the new statement and if everything is ok.


fietew:
Admitting, that I have only checked a subset of the files, this is ready for merge. For non-merged pull-requests, it would be nice to provide the script for the renaming in order to change the license information for the new files in the respective branches.

hagenw:
You can find them here: https://github.com/sfstoolbox/maintenance-tools

You have to go to the folder and execute something like
`for FILE in *.m; do change_copyright_to_mit.sh $FILE; done`
and make sure that `sfs-copyright-mit.txt`is available.

Issue #67, opened 2016-02-29
Speed up SOFA


hagenw:
As @fietew wrote in #45:

> As far as I understood the current implementation, the SOFA file is accessed for each head orientation and for each loudspeaker position separately. This means, that an impulse response (having e.g. the index "42" ) is loaded from the hard disk drive into the RAM several times. If we really want to speed up things, then this architecture has to be reassessed. An alternative workflow could be:

> * iterate over the head orientations and the secondary source position in order to get the indices of the impulse responses, which are needed from the SOFA file. This can be done by evaluating the header of the SOFA file, i.e. using nearest_neighbor, etc. (as you have done it in the current implementation)
* the indices are likely to be non-unique, as some head-orientations-ssd-position pairs might need the same impulse responses. However, functions like ``unique`` (http://de.mathworks.com/help/matlab/ref/unique.html) help us to remove those duplicates.
* sorting the indices might also help to identify segments of connected indices, i.e. groups of directly subsequent indices, e.g. [2,3,4,5] [7,8,9] [11,12]. As I experienced, that loading the impulse responses index by index is quite slow in SOFA, one could to something like https://github.com/TWOEARS/binaural-simulator/blob/master/src/%2Bsimulator/DirectionalIR.m#L189-L202 .
* after loading the impulse responses, they have to be processed according to the head-orientations-ssd-position pairs they are need for, i.e. distance adjustment, angle interpolation, adding up all loudspeakers.

> The major drawback of this approach is, that there might be a huge number of impulse responses which are loaded into the RAM at once. One could define a maximum number of irs loaded in one step and then iterate as long as there are impulse responses left to be processed. However, one would have ensure, that there is always at least one head-orientations-ssd-position pair which can be processed with the current impulse responses in the RAM.

If someone has an idea how to do this or some time left we can work on this, but the priority is not very high at the moment.

PR #90, opened 2016-06-09
Remove conf.N setting

hagenw:
This resulted from a discussion in #89.

The idea is that the user don't have to specify the desired impulse response link with `conf.N`, but that the Toolbox automatically decides what is the best length.

This would be beneficial, especially for new users. The only downside I see at the moment is that the resulting impulse responses could be very long if the original data from the SOFA file is long.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -5,6 +5,7 @@
     - update delayline() config settings
     - add Lagrange and Thiran filters to delayline()
     - replace wavread and warwrite by audioread and savewav
+    - add new online doc at http://matlab.sfstoolbox.org/
 
 2.1.0 (10. March 2016)
     - make conf struct mandatory
--- a/NEWS
+++ b/NEWS
@@ -5,6 +5,8 @@
     - update delayline() config settings
     - add Lagrange and Thiran filters to delayline()
     - replace wavread and warwrite by audioread and savewav
+    - convolution() excepts now two matrices as input
+    - allow headphone compensation filter to be a one- or two-channel wav file
     - add new online doc at http://matlab.sfstoolbox.org/
 
 2.1.0 (10. March 2016)
--- a/README.rst
+++ b/README.rst
@@ -104,9 +104,10 @@ can look like this:
     conf.fs = 48000;
 
 For a detailed description of all available features the SFS Toolbox, have a
-look at the :ref:`sec-feature-documentation`.
+look at the `feature-documentation`_.
 
 .. _online documentation: http://matlab.sfstoolbox.org
+.. _feature documentation: http://matlab.sfstoolbox.org/en/latest/feature-documentation/
 
 
 Credits and feedback
--- a/README.rst
+++ b/README.rst
@@ -104,7 +104,7 @@ can look like this:
     conf.fs = 48000;
 
 For a detailed description of all available features the SFS Toolbox, have a
-look at the `feature-documentation`_.
+look at the `feature documentation`_.
 
 .. _online documentation: http://matlab.sfstoolbox.org
 .. _feature documentation: http://matlab.sfstoolbox.org/en/latest/feature-documentation/
--- a/SFS_ir/ir_correct_distance.m
+++ b/SFS_ir/ir_correct_distance.m
@@ -66,7 +66,6 @@ delay = (r-ir_distance)/c*fs; % / samples
 % This gives weight=1 for r==ir_distance
 weight = ir_distance./r;
 % Make sure impulse response is long enough for the specified delay
-ir = cat(3,ir,zeros(size(ir,1),size(ir,2),delay+offset));
+ir = cat(3,ir,zeros(size(ir,1),size(ir,2),max(ceil(delay+offset))));
 % Apply delay and weighting
-ir = delayline(ir,[delay+zero_padding; delay+zero_padding], ...
-               [weight; weight],conf);
+ir = delayline(ir,[delay+offset; delay+offset],[weight; weight],conf);
--- a/SFS_ir/ir_correct_distance.m
+++ b/SFS_ir/ir_correct_distance.m
@@ -60,13 +60,6 @@ N = conf.N;
 
 
 %% ===== Computation ====================================================
-% Stop extrapolation for distances larger than 10m
-if any(ir_distance>10)
-    ir_distance = min(ir_distance,10);
-    warning(['%s: Your desired radius is larger than 10m, but we will ', ...
-        'only extrapolate up to 10m. All larger radii will be set to ', ...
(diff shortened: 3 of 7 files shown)


hagenw:
@VeraE: I have updated `ir_correct_distance()` and removed `conf.N` there. Could you have a look if this is still ok, please.

hagenw:
I started working on removing `conf.N` from the time-domain driving functions. For WFS it should work now.

@VeraE, @fietew: for `wfs_preequalization()` and `driving_function_imp_nfchoa()` I'm not sure how to do it.
1.) For `driving_function_imp_nfchoa()` the size is specified at [line 88](https://github.com/sfstoolbox/sfs/blob/remove_N/SFS_time_domain/driving_function_imp_nfchoa.m#L88) and it will not work if there are no zeros padded at the end. But I don't know how to calculate how many zeros we need.
2.) For `wfs_preequalization()` zeros are added at [line 83](https://github.com/sfstoolbox/sfs/blob/remove_N/SFS_time_domain/wfs_preequalization.m#L83) for IIR filtering, is this needed and how we can replace it?

@fietew: could you have a look at `driving_function_imp_localwfs()` and `driving_function_imp_wfs_vss()` and see if you can replace `conf.N` there in a similar way I have done it for `driving_function_imp_wfs()`?

VeraE:
I did some updates on `ir_correct_distance()`.
Now there remains the problem that in `ir_generic()` the impulse responses that are delivered by `get_ir()` could have different lengths but should be saved in one array. At the moment the impulse responses are first convoluted and then truncated to conf.N, so calculations finish succesfully. Getting rid of the conf.N restriction here would mean to either pre-calculate the maximum delay or adapt the size of the array in every loop over x0.  I fear both options would increase computation time. Which do you think we should try first?

hagenw:
Mh, too bad, this does not sound like there is an easy solution.
I'm a little bit afraid of your second approach and changing sizes inside a loop is normally a bad idea in Matlab. The first one would require something very similar to #67, which also would require a pre-scanning of the needed impulse responses. So far I was not very motivated to do this as I was afraid that the code of `ir_generic()` would not be easy to understand afterwards.

Do you still think it is worse trying to remove `conf.N` also from `ir_generic()`?
In my opinion it would be ok, to only use it there (and maybe rename it). In this way we would also maintain the nice possibility of shortening the impulse response in `ir_generic()`.

VeraE:
You are right both approaches are not a good idea.

To me it doesn't make sense to only remove conf.N in some parts of the calculation, though. As long as there is one step where impulse responses are shortened the user has to know before how to set conf.N sensibly. I was hoping we could come up with a solution that would not make the code much more complicated or slower. Unfortunately, I couldn't think of a better way to implement this so far. So maybe, we have to drop this completely?

hagenw:
For the impulse responses part we could still think about if it is a good idea to not include it into `get_ir()` and `ir_correct_distance()` or if we should stay with the old approach. For me both would be fine. At the current master version `get_ir()` returns already a length of `conf.N` (only for the HRIR case, for BRIR we have to modify it).

But I'm still very interested in getting rid of `conf.N` from all the time-domain driving functions. There should be no need for the user to specify it there.

VeraE:
Fiete has just helped me to fix this branch as I managed to mess up the three commits you did before my merge. 

So your idea would be to have truncating to conf.N at a higher level in the code? We can do that. So I'd leave `ir_correct_distance()` as it now in this branch and have a look at the commits done by you for the driving functions.

VeraE:
Time-domain driving function for WFS looks fine. I just got one question: why is the impulse response zero-padded by `delay_offset` as well in [line 140](https://github.com/sfstoolbox/sfs/blob/remove_N/SFS_time_domain/driving_function_imp_wfs.m#L140) of `driving_function_imp_wfs()`? I'd think this is unnecessary but maybe I am missing something here...?

The WFS IIR pre-equalization does not work properly without the zero-padding as the Matlab function `sosfilt()` in `wfs_prequalization()` is returning a signal with the same length as its input signal (which is 1 when called from `driving_function_imp_wfs()`). We would have to specify a new parameter here if conf.N is not used, I fear.

I had a look at `driving_function_imp_nfchoa()` too, but have no idea how to precalculate the zeros either at the moment. I've never used this function extensively.

hagenw:
With the `delay_offset` you are probably right. I was not sure, so I have been gone for the safe choice.

As it also seems complicated to remove `conf.N` from the driving functions, we maybe give up on it altogether. I will wait for @fietew opinion.

fietew:
> @VeraE, @fietew: for wfs_preequalization() and driving_function_imp_nfchoa() I'm not sure how to do it. 
1.) For driving_function_imp_nfchoa() the size is specified at line 88 and it will not work if there are no zeros padded at the end. But I don't know how to calculate how many zeros we need.
2.) For wfs_preequalization() zeros are added at line 83 for IIR filtering, is this needed and how we can replace it?

It's an IIR-Filter. You theoretically need an infinite number of zeros. Maybe we could truncate the impulse responses if the power/amplitude of the last coefficient is below a given threshold.

> @fietew: could you have a look at driving_function_imp_localwfs() and driving_function_imp_wfs_vss() and see if you can replace conf.N there in a similar way I have done it for driving_function_imp_wfs()?

There is no ``driving_function_imp_wfs_vss()``, btw. I think the current implementation does not take the delay offset caused by the fractional delay filter into account. I necessary, I could implement a function which returns just the delay of the chosen fractional delay method.

hagenw:
@fietew:
Here you can find [``driving_function_imp_wfs_vss()``](https://github.com/sfstoolbox/sfs/blob/remove_N/SFS_time_domain/driving_functions_imp/driving_function_imp_wfs_vss.m).
At line [64](https://github.com/sfstoolbox/sfs/blob/remove_N/SFS_time_domain/driving_functions_imp/driving_function_imp_wfs_vss.m#L64) ``conf.N`` is used. And in [``driving_function_imp_localwfs()``](https://github.com/sfstoolbox/sfs/blob/remove_N/SFS_time_domain/driving_function_imp_localwfs.m) at line [84](https://github.com/sfstoolbox/sfs/blob/remove_N/SFS_time_domain/driving_function_imp_localwfs.m#L84).

And do we have to account for the delay offset of the fractional delay filter for specifying the driving signal length? For the normal WFS case I'm considering now only the maximum delay without possible delay offsets.

fietew:
I discussed with @trettberg and we somehow agreed, that keeping ``conf.N`` is a more consistent way than perhaps dealing with driving signal of different lengths. Imho this can be closed.

hagenw:
OK, I agree.

PR #37, opened 2015-09-15
Corrected Driving Functions for 2.5D WFS

fietew:
This branch is still work in progress. I added the stationary phase approximated driving functions (reference point and reference line) for point sources, focused source and plane waves. Furthermore, I tried to remove doubled driving functions.
The time domain driving functions for plane waves are still missing.

PR #96, opened 2016-06-15
Update links to online documentation

hagenw:


hagenw:
The links in the WFS driving functions are corrected in #37.

Issue #76, opened 2016-03-21
HRIR interpolation point selection


hagenw:
I'm not sure if the simple usage of nearest neighbor points for performing the interpolation is the best approach for impulse responses. Here is an easy example:

```Matlab
x0 = [45 46 46 50; 0 -2 2 0];
xs = [47; 0];
x0_int = findnearestneighbour(x0,xs,3);
figure; plot(x0(1,:),x0(2,:),'xb')
axis square; axis([43 52 -3 3]); hold on
plot(xs(1),xs(2),'xr')
plot(x0_int(1,:),x0_int(2,:),'or','MarkerSize',10)
xlabel('Azimuth / deg')
ylabel('Elevation / deg')
```

![ir_interpolation](https://cloud.githubusercontent.com/assets/173624/13924060/27c04cee-ef83-11e5-81ce-11556b91cfb2.png)

As you can see we want to get the point shown as red cross. The red circles are the three nearest neighbors that are selected for the interpolation. Afterwards it is checked if the nearest and second nearest neighbor have the same azimuth or elevation. If so only these points are used. In this example all three points will be used for interpolation.
The problem is that in this case it would in my opinion make more sense to use just the two points that have both an elevation of 0°. The problem is I don't know if we are able to write an algorithm that does this sort of clever selection that works with every grid and point spacing.
Has anyone some experience on this?

fietew:
Maybe this is interesting:
https://en.wikipedia.org/wiki/Natural_neighbor
the German version explains it a little bit better:
https://de.wikipedia.org/wiki/Voronoi-Interpolation

fietew:
Voronoi regions for a sphere https://github.com/tylerjereddy/py_sphere_Voronoi

hagenw:
Cool, especially the second link.

Now, we only need a person who is going to implement this.

trettberg:
Pushed some code to a branch named vbap_style_interpolation.

Here is the example from above.
(The upper point actually gets a weight of zero. Just the markersize has an offset.)

```Matlab
azi = [0; 1; 1; 5;]*pi/180;
elev = [0; 1 ; -1; 0;]*pi/180; 
[x,y,z] = sph2cart(azi,elev,ones(4,1));
x0 = [x y z];
[x,y,z] = sph2cart(2*pi/180,0,1);
xs = [x y z];

[indeces,weights] = findconvexcone(x0,xs);
x0_int = x0(indeces,:);

[...]

PR #101, opened 2016-06-27
Interpolation toa

VeraE:
I propose a different interpolation method for HRIRs that calculates a linear interpolation in the frequency domain for magnitude and phase separately. This way, the time of arrival in the HRIRs is interpolated as it is to be expected.
At the moment I added a new conf-parameter called `conf.ir.interpolationmethod` that can be `'simple'` (= the old interpolation method samplewise in the time domain without considering the time of arrival) or `'freqdomain'`.
As the new method makes use of the FFT symmetries for real signals and interpolates only one half of the spectrum, it is only a bit slower than the old method (about 2-3% on my system). Should we still keep the old method?
I could extend the new method to 2D interpolation as well.


fietew:
1. Does the ``unwrap`` of the phase really solve all the issues with the linear interpolation of the phase, i.e. does it consider all the "wrap around" effects w.r.t 2*pi?
2. Maybe a reference for the algorithm should be added to the comments

fs446:
maybe this is inspiring:

http://iem.kug.ac.at/projects/workspace/2011/phasenabrollung-zur-interpolation-von-aussenohruebertragungsfunktionen-auf-der-kugel.html

VeraE:
Thanks for your comments. I updated the interpolation so that phase aliasing is avoided and added references.
The 2D interpolation for our case should work the same way but I would postpone that to when #76 is finished as I think this could affect the parameters that `interpolate_ir()` receives and possibly make the distinction in 1D and 2D unnecessary.
Should the warning for backwards compatibility better be turned off?
If this looks alright to you, I'll update the comment header.

hagenw:
Looks good, thanks.
Could you maybe create a function `test_interpolation.m` in the `validation` folder and add some code that compares the outcome of `'simple'` with `'freqdomain'`. Would be cool to have a nice comparison.

hagenw:
In `SFS_config.m` you are referencing the file `test_interpolation_methods.m` which should be placed in the `validation` folder, but this file does not exist. Could you add one?

VeraE:
Sorry for the delay, I had to wait for publishing of the HRTFs with low frequency correction with a new license before I could finish the test file. I have added it now. Does this look alright to you?

hagenw:
Cool, this looks nice now and I'm going to merge it.

PR #103, opened 2016-07-19
replace custom fft, ifft functions with built-in functionality

fietew:
Things are quite confusing with ``fft_real`` and ``ifft_real``, as they do not provide any additional functionality which is not provided by MATLAB's ``fft`` and ``ifft``. Furthermore, convolution can be improved, if ``x`` is a vector and ``y`` is a matrix (or vice versa), as the FFT has to be only performed once on the vector.


VeraE:
I tested the new `convolution()` function and it works nicely. To me, the code is now more obvious, too.

Issue #108, opened 2016-08-23
Bug: Inconsistent delay_offset computation in time domain wfs driving functions


fietew:
The `delay_offset` return value of the `delayline` is given in samples, while the calculations in `driving_function_imp_wfs` are assuming seconds. Which convention is the correct one?

hagenw:
Both are correct. The whole calculation in `delayline()` is done in samples as I found this to be more intuitive when working with integer delays and the whole calculation for the WFS driving functions is performed in seconds.
But you are correct, there is a bug then in [line 150 of `driving_function_imp_wfs`](https://github.com/sfstoolbox/sfs-matlab/blob/master/SFS_time_domain/driving_function_imp_wfs.m#L150) which reads
```Matlab
delay_offset = delay_offset + delayline_delay;
```
at the moment and should read
```Matlab
delay_offset = delay_offset + delayline_delay/fs;
```
instead.

Maybe we should avoid this by restricting us to use always t / s as input and output arguments in the whole Toolbox? At the moment `sound_field_imp()` uses also t / samples.

fietew:
Yeah, as the purpose of ``delay_offset`` is to somehow incorporate its value into ``sound_field_imp()`` in order to get the correct time snapshot, it is quite complicated that one has to convert it from seconds back to samples. However, I don't know how many functions would be effected from such a change and if it's worth the effort.

fietew:
Maybe, we should merge the fix and open another issue/pull request to discuss the seconds/samples convention?

hagenw:
OK, I merged it and created a new branch `time_in_sec` where we can start work and discuss a change in the general time behavior (no pull request yet as this is not possible without a change in code).

PR #109, opened 2016-08-23
Make handling of t consistent

hagenw:
The time `t` is handled in some functions in samples and in others in seconds. To avoid confusions like the one of #108, we should make this more consistent. My proposal would be to switch completely to t in seconds as we also do in the Python version.

So, let's start with a list of all the functions that uses t in samples which should be reviewed if we easily could change them:
- [x] `delayline()`
- [x] `warning_if_zero()`
- [x] `greens_function_imp()`
- [x] `sound_field_imp_line_source()`
- [x] `sound_field_imp_localwfs()`
- [x] `sound_field_imp()`
- [x] `sound_field_imp_nfchoa()`
- [x] `sound_field_imp_plane_wave()`
- [x] `sound_field_imp_point_source()`
- [x] `sound_field_imp_wfs()`
- [x] `time_response_wfs()`
- [x] `time_response_localwfs()`
- [x] `time_response_nfchoa()`

In addition we have a few functions that use samples, but in those cases it is not directly referring to the time t, but to something like filter length. For those I would propose to stay with samples, but we should also review them:
- [x] `dummy_irs()`
- [x] `general_least_squares()`
- [x] `hann_window()`

As a first step it would be great to check, if I have missed some function in the listing.


validation/test_delayline.m
```diff
@@ -55,12 +55,11 @@
 
 %% ===== Configuration ==================================================
 % Delays to evaluate
```
chris-hld: These delay values are indeed samples.

hagenw:
I have no updated all functions I was able to identify that used t in samples to use t in seconds instead.
for the user the biggest change will be in the `sound_field_imp` related functions as there t must be provided. What will be much easier now is the incorporation of the returned `delay_offset` values as they now always are in seconds and can directly be applied.

The three functions I listed above that still use samples instead of seconds should stay with this in my opinion.

@fietew: could you please have a look at my changes.

fietew:
Just a short note about the handling of this changes: As @hagenw already mentioned, this pull request will have quite an impact on the behavior of the time-domain functionalities. I would suggest to create a minor release (2.2.x) right before **and** right after merging this pull request. This enables the correct referencing of the used implementation, e.g. in scientific publications.

hagenw:
OK, I agree, we should first make a release, then integrate and then make another release.

SFS_general/delayline.m
```diff
@@ -95,112 +98,114 @@
...
 
 % --- Expand dt and weight ---
 % If only single valued time delay and weight is given, create vectors
 if channels>1 && length(dt)==1, dt=repmat(dt,[1 channels]); end
 if channels>1 && length(weight)==1, weight=repmat(weight,[1 channels]); end
 
+
 %% ===== Conversion to integer delay =====================================
-dt = rfactor.*dt;  % resampled delays
+dt = rfactor.*dt.*fs;  % resampled delays / s
```
fietew: Regarding the comment: It is actually in samples now, isn't it?

SFS_general/delayline.m
```diff
@@ -95,112 +98,114 @@
...
+    rfactor = 1.0;
+    delay_offset = 0.0;
+case 'matlab'
+    rfactor = delay.resamplingfactor;
+    delay_offset = 0.0;
+    sig = resample(sig,rfactor,1);
+case 'pm'
+    % === Parks-McClellan linear phase FIR filter ===
+    rfactor = delay.resamplingfactor;
+    delay_offset = (delay.resamplingorder*rfactor/2) / fs;
```
fietew: It might be better to leave the ``delay_offset`` in samples during the computations and convert into seconds at the end of the function.

fietew:
One additional thing: Under MATLAB, I had problems to execute ``test_plot(true)``:
```
Attempt to execute SCRIPT test_plot as a function:
/home/winter/projects/sfstoolbox/test_plot.m
```

trying ``test_plot`` without argument got me this single plot
untitled

After I switched to the ``validation`` directory, everything worked as expected. If think, there is some kind of naming issue. Can somebody else test the function under MATLAB?

SFS_general/delayline.m
fietew: Maybe change the order to ``dt.*rfactor.*fs``

SFS_general/delayline.m
hagenw: I guess this is better as we avoid lots of `/ fs`. Originally I thought it might be a good idea to use always time, but `dt` is also converted to samples as you pointed out with the next comment.
So, I changed the behavior of `delay_offset` which is now converted into time only at the end of the function.

SFS_general/delayline.m
hagenw: I corrected this, thanks.

hagenw:
@chris-hld could you please verify if you can reproduce the `test_plot` problem under Matlab.
Maybe we should simply rename that function.

fietew:
> Maybe we should simply rename that function.
I agree.

chris-hld:
No, I cannot recreate that problem.
Executing:
`SFS_start; test_plot()` ,
results in correct `Not enough input arguments.`.
Is there maybe another test_plot()  in your path?

fietew:
Let's just rename the function in order to finally merge this pull request.

hagenw:
I renamed the function to `test_plotting()`.
Unfortunately we are not ready for merging as the documentation (files under `doc/`) has to be adjusted. I will have a look at it.

hagenw:
I have updated the documentation.

@chris-hld could you please run `generate_plots` and push only the following files afterwards:
```
sound_field_imp_nfchoa_25d.png
sound_field_imp_nfchoa_25d_dB.png
sound_field_imp_multiple_sources_dB.png
```

@fietew could you please run the `test_all` a last time to see if still everything is working.

After that we should be ready to merge.

fietew:
``test_all`` finished without significant errors (beside ``to_be_implemented``). 

Looking at the plots:
``test_colormaps``: In the time-domain plots, there is no soundfield visible. Wrong d

hagenw:
Thanks, there was indeed one time given in samples left. I have fixed `test_colormaps()` now.

chris-hld:
That should be it

hagenw:
I have cleaned up the history of this and included it into master as seven different commits.

PR #114, opened 2016-11-22
add secondary source selection and tapering for WFS auralisation

VeraE:
Secondary source selection and tapering had been missing for WFS auralisation and are now added in `ssr_brs_wfs()`.


hagenw:
Oh, that's interesting. I guess I have to check the scripts from my last experiments.

PR #115, opened 2016-12-01
Correct WFS driving function time delay

hagenw:
This is a follow up on #113 and deals only with the delay of the impulses.

Try the following code:
```matlab
conf = SFS_config;
X = [0 0 0];
phi = 0;
xs = [2.5 0 0];
src = 'ps';
hrtf = dummy_irs(conf);
% Using an impulse response
ir = ir_wfs(X,phi,xs,src,hrtf,conf);
% Using sound_field_imp() for every time step
[s,t] = time_response_wfs(X,xs,src,conf);
figure;
plot(0:700,ir(1:701,:),'-g');
hold on
plot(t,s,'-b');
grid;
xlabel('samples');
hold off;
```

Before the result was

current

The desired time delay of the first wave front is 321 samples corresponding to a distance of 2.5 m of the virtual point source. The blue impulse response created with `time_response_wfs()` is shifted to the left by the delay offset from the WFS driving function due to the time reversal in `sound_field_imp()`. The green impulse response is shifted to the right by the same delay. This is not a new discovery, see #62, but we have corrected it only [inside `sound_field_imp_wfs()`](https://github.com/sfstoolbox/sfs-matlab/blob/master/SFS_time_domain/sound_field_imp_wfs.m#L100) so far.

This pull request now also corrects it in `time_response_wfs()` and `ir_wfs()`. Running the code from above now results in

fixed

@VeraE, @fietew: is it desireable to fix this delay at all inside `ir_wfs()`? Or would it be better to add it two times inside `time_response_wfs()` to correct for the difference?

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,7 +1,10 @@
 2.3.0 ()
-    - fix Gauss-Legendre quadrature weights
-    - fix default 2.5D WFS driving function in time domain
+    - rename conf.wfs.t0 to conf.t0
+    - add delay_offset as return value to NFC-HOA time domain driving funtions
     - correct absolute amplitude values of WFS in time domain
+    - fix default 2.5D WFS driving function in time domain
+    - fix missing secondary source selection in ssr_brs_wfs()
+    - fix Gauss-Legendre quadrature weights
 
 2.2.1 (22. August 2016)
     - fix delayoffset for FIR fractional delay filter
--- a/SFS_analysis/time_response_wfs.m
+++ b/SFS_analysis/time_response_wfs.m
@@ -88,14 +88,15 @@ x0 = secondary_source_tapering(x0,conf);
 % Generate time axis (0-500 samples)
 t = (0:500)';
 s = zeros(1,length(t));
-d = driving_function_imp_wfs(x0,xs,src,conf);
+% t0 is a time offset (delay) added by the driving function
+[d,~,~,t0] = driving_function_imp_wfs(x0,xs,src,conf);
 % If desired a cosine shaped pulse instead of the default dirac pulse could be
 % used
 % d = convolution(d,hann_window(5,5,10));
 for ii = 1:length(t)
     if showprogress, progress_bar(ii,length(t)); end
     % calculate sound field at the listener position
-    p = sound_field_imp(X(1),X(2),X(3),x0,greens_function,d,t(ii),conf);
+    p = sound_field_imp(X(1),X(2),X(3),x0,greens_function,d,t(ii)+t0*fs,conf);
     s(ii) = real(p);
 end
 
--- a/SFS_binaural_synthesis/ir_nfchoa.m
+++ b/SFS_binaural_synthesis/ir_nfchoa.m
@@ -1,7 +1,7 @@
-function [ir,x0] = ir_nfchoa(X,phi,xs,src,sofa,conf)
+function [ir,x0,delay] = ir_nfchoa(X,phi,xs,src,sofa,conf)
 %IR_NFCHOA generates a binaural simulation of NFCHOA
 %
-%   Usage: [ir,x0] = ir_nfchoa(X,phi,xs,src,sofa,conf)
+%   Usage: [ir,x0,delay] = ir_nfchoa(X,phi,xs,src,sofa,conf)
 %
 %   Input parameters:
 %       X       - listener position / m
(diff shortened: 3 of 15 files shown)


hagenw:
I was wrong. `conf.wfs.t0` is set per default to `system`, so the expected delay is 192 corresponding to 1.5m to the first active loudspeaker. In order to fix this I had to adjust `dummy_irs()` to have the dirac at the first entry as the correct distance is always added by `get_ir()` as we decided in #92. I don't know if this is the best way to go, but at least at the moment the correct samples can be simulated for `conf.wfs.t0` set to `system` and to `source`.

Here is the result from the code above:

fix_time_delay1

fietew:
I think we are in a dilemma between having the correct delay corresponding to the virtual sound field and keeping all the information from the impulse response. If `X` gets close to the SSD, we will cut off the impulse response, see:

```matlab
conf = SFS_config;
X = [1.45 0 0];
phi = 0;
xs = [2.5 0 0];
src = 'ps';
hrtf = dummy_irs(conf);
% Using an impulse response
ir = ir_wfs(X,phi,xs,src,hrtf,conf);
% Using sound_field_imp() for every time step
[s,t] = time_response_wfs(X,xs,src,conf);
figure;
plot(0:700,ir(1:701,:),'-g');
hold on
plot(t,s,'-b');
grid;
xlabel('samples');
hold off;
```

VeraE:
I had a closer look at the delays, especially in `ir_wfs()`. Apart from the problem that Fiete pointed out, I noticed that `delayline()` gets called 3 times in `ir_wfs()`:

1. when calculating the driving function,
2. in `ir_correct_distance()` and
3. when compensating the delay of the driving function from point 1. (consisting of prefilter delay, compensating for the case `conf.wfs.t0 = 'source'` and optional delay of resampling and/or fractional delay method in `delayline()`).

Point 2. could be corrected. But depending on the resampling and/or fractional delay method, the resulting impulse response cannot be free of "processing delays" because of point 3. In combination with the point raised by Fiete, it might be better to not correct for the delays at all? Instead they could be just returned to let the user know so he can understand the start of the simulated impulse response.

The auralisation function `ssr_brs_wfs()` would have to be treated in the same way as `ir_wfs()` I think.

I think the change to `dummy_irs()` in bfb4f0c is a good thing, though, because it leads to the same handling of the impulse response as for measured HRTFs.

hagenw:
@fietew: that's a good point. For `time_response_wfs()` I don't see a problem with this, but for `ir_wfs()` it might be indeed preferable to avoid cutting into the signal as this can happen without the knowledge of the user.

@VeraE: that's another good point: depending on the exact applied delay line method, there could always be a delay. Your point of calling `delayline()` to often is handled in an extra issue: #116.

I would propose to do the following:
1. Correct the delay in `time_response_wfs()` to match the [behavior of `sound_field_imp_wfs()`](https://github.com/sfstoolbox/sfs-matlab/blob/master/SFS_time_domain/sound_field_imp_wfs.m#L100)
2. Add the delay offset as a third return parameter to `ir_wfs()`
3. Stay with the changes to `dummy_irs()`

The main downside of this solution is that `ir_wfs()` and `time_response_wfs()` will provide different results, but I guess we can handle this by an entry to the documentation.

VeraE:
I agree with this. Thanks for applying the fixes. As I mentioned above, `ssr_brs_wfs()` should be treated in the same way as `ir_wfs()`: that means `driving_function_imp_wfs()` should return its delay offset and `ssr_brs_wfs()` should return the overall delay to the user. Could you incorporate this as well? Or I can do this in the end, too.

`ssr_brs_wfs()` also has the problem of calling `delayline()` more than once (for calculation of the driving function and for `ir_correct_distance()` just as in `ir_wfs()`), I will add this to #116.

hagenw:
I added `delay` as an additional return parameter to `ir_wfs()` and `ssr_brs_wfs()`. All those delay stuff does of course also apply tho NFC-HOA, but this is not integrated yet and will be handled at #95. After finishing this one, I will include the changes there for NFC-HOA as well.

Now you can do the following to get an aligned plot:
```Matlab
conf = SFS_config;
X = [0 0 0];
phi = 0;
xs = [2.5 0 0];
src = 'ps';
hrtf = dummy_irs(conf);
[ir,~,delay] = ir_wfs(X,phi,xs,src,hrtf,conf);
offset = round(delay*conf.fs);
[s,t] = time_response_wfs(X,xs,src,conf);
figure;
plot(0:700,ir(1+offset:701+offset,:),'-g');
hold on
plot(t,s,'-b');
grid;
xlabel('samples');
hold off;
```

NOTE: there is still an offset of 1 sample between bot impulse responses, which could come from a slightly different use of time axis in both cases.

hagenw:
The commits are now integrated into master, besides the last one regarding documentation, this will be continued at #118.

PR #118, opened 2016-12-07
DOC: include new section on system impulse responses

hagenw:
This should update the section on binaural simulations in the docs, reflecting the newest changes we did.


hagenw:
Thanks, for `doc/img/sound_field_imp_nfchoa_25d.png` there seems to be a problem with the x- and ylabel, which is has now not enough space to the tics.

chris-hld:
There was an unpredictable behavior related to the switching of font. 
`pause(0.2)` around the png rendering helped.

hagenw:
Thanks, looks nice.

Is there a difference between OpenGl 4.4.0 and 4.5.0. At a first glance the images look identical to me.

hagenw:
I don't know if this comes from the `axis([-2 2 -2 2]);`, but some of the plots have now xtics at -2, 0, 2, whereas the ytics are at -2, -1, 0, 1, 2, which is the desired behavior. If you could fix this and replot ones more it should be fine and I will start to remove all the unnecessary commits to save space.

chris-hld:
I suppose not -- hard to tell since the png resolution is quite low anyway

chris-hld:
This also comes from the rescaling related to font switching. I think these had been like this before. I will introduce hard axis ticks.

hagenw:
Continued at #134.

PR #119, opened 2016-12-07
Update usage of bandpass in sound_field_imp()

hagenw:
This is a follow up of #113.

At the moment we have the default config setting:
```Matlab
conf.usebandpass = true;
conf.bandpassflow = 10;
conf.bandpassfhigh = 20000;
```
This setting results in an execution of the following line inside `sound_field_imp()`:
```Matlab
d = bandpass(d,bandpassflow,bandpassfhigh,conf);
```
in `bandpass()` the function `fir2()` is used to create the filter and applied by `convolution()`.

`sound_field_imp()` is the only place where the bandpass filtering is applied at the moment.

This pull request should figure out, why we have enabled the bandpass filtering as the **default setting** and if we can disable as default or even remove it.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,4 +1,5 @@
 2.3.0 ()
+    - change default value of conf.usebandpass to false
     - rename conf.wfs.t0 to conf.t0
     - add delay_offset as return value to NFC-HOA time domain driving funtions
     - correct absolute amplitude values of WFS in time domain
--- a/SFS_config.m
+++ b/SFS_config.m
@@ -97,10 +97,6 @@ conf.showprogress = false; % boolean
 conf.fs = 44100; % / Hz
 % Speed of sound
 conf.c = 343; % / m/s
-% Bandpass filter applied in sound_field_imp()
-conf.usebandpass = true; % boolean
-conf.bandpassflow = 10; % / Hz
-conf.bandpassfhigh = 20000; % / Hz
 
 
 %% ===== Delayline =======================================================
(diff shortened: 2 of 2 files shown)


hagenw:
If we simulate an impulse response at a single listening position, as for example done by `time_response_wfs()` the bandpass filtering results in a reduced amplitude as shown in #113.

Here, I do sound field simulations in the time domain for WFS and NFC-HOA with and without bandpass filtering.

**WFS**
```Matlab
conf = SFS_config;
conf.plot.usedb = true;
conf.usebandpass = true;
sound_field_imp_wfs([-2 2],[-2 2],0,[0 2.5 0],'ps',200,conf)
conf.usebandpass = false;
sound_field_imp_wfs([-2 2],[-2 2],0,[0 2.5 0],'ps',200,conf)
```
wfs_w_bandpass
wfs_wo_bandpass

**NFC-HOA**
```Matlab
conf = SFS_config;
conf.plot.usedb = true;
conf.usebandpass = true;
sound_field_imp_nfchoa([-2 2],[-2 2],0,[0 2.5 0],'ps',200,conf)
conf.usebandpass = false;
sound_field_imp_nfchoa([-2 2],[-2 2],0,[0 2.5 0],'ps',200,conf)
```
nfchoa_w_bandpass
nfchoa_wo_bandpass

Looking at both results I would say that both looks better without the bandpass filtering.

@fietew: do you have an idea for an example that might look better with bandpass filtering?
@spors: if I remember the bandpass filtering was included from the beginning in the time-domain scripts, could you comment on the idea behind it

hagenw:
@fietew: would it be ok with you, if I change the default value of `conf.usebandpass` to `false`?

fietew:
Is it used somewhere else appart from the time-domain sound field computation?

hagenw:
No.

fietew:
I will have a quick chat with @spors, whether i am missing any benefit of using the bandpass filtering. I don't see any.

hagenw:
Ok, thanks.

fietew:
One meaningful application could be, that the user wants to simulate the sound field using an bandlimited impulse ranging e.g. from 0 to 1 kHz instead of a fullband impulse. However, switching the default of ``conf.usebandpass`` to ``false`` make sense imho.

PR #124, opened 2016-12-21
Correct g0 for 2.5D WFS point source and focussed source

hagenw:
Looking at the definition of g0 for a point source [using a reference point](http://sfstoolbox.org/en/latest/#equation-D.wfs.ps.2.5D) and [using a reference line](http://sfstoolbox.org/en/latest/#equation-D.wfs.ps.2.5D.refline) and comparing those to the definitions provided in Start (1997) eq. (3.16) or Verheijen (1997) eq. (A.14) there seems to be a mismatched in the sign:
```Latex
\sqrt(\frac{\xref}{\xref-\xs})
```
vs.
```Latex
\sqrt(\frac{\xref}{\xref+\xs})
```

This pull request changes our implementation to be in line with the documentation at http://sfstoolbox.org, which is in line with Schultz (2016).

@fs446 could you please have a look, if this pull request is the correct implementation or if the implementation before was correct, but not the documentation?

### DIFF
--- a/SFS_monochromatic/driving_functions_mono/driving_function_mono_wfs_fs.m
+++ b/SFS_monochromatic/driving_functions_mono/driving_function_mono_wfs_fs.m
@@ -179,14 +179,12 @@ elseif strcmp('2.5D',dimension)
         % r = |x0-xs|
         r = vector_norm(x0-xs,2);
         % 2.5D correction factor
-        %         _____________________
-        %        |      |xref-x0|
-        % g0 = _ |---------------------
-        %       \|||xref-x0| - |xs-x0||
+        %         ___________________
+        %        |     |xref-x0|
+        % g0 = _ |-------------------
+        %       \||xref-x0| + |xs-x0|
         %
-        % Verheijen (1997), eq. (A.14)
-        %
-        g0 = sqrt( vector_norm(xref-x0,2) ./ abs(vector_norm(x0-xref,2) - r) );
+        g0 = sqrt( vector_norm(xref-x0,2) ./ abs(vector_norm(x0-xref,2) + r) );
         %                       ___     ___
         %                      | 1     |-iw  (xs-x0) nx0
         % D_2.5D(x0,w) = g0  _ |---  _ |--- ------------- e^(i w/c |x0-xs|)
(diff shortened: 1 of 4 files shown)


fietew:
@hagenw: Please use the test function ``test_wfs_25d`` in order to check, if the changed driving functions produce a correct amplitudes. I feel, that they the driving functions are not correct anymore. The variables in the code do not necessarily represent the variables of the formulas. An example: For the WFS 2.5D driving function for a point source (reference line): 

```matlab
% Distance ref-line to linear ssd
dref = vector_product(xref-x0,nx0,2);
% Distance source and linear ssd
ds = vector_product(xs-x0,nx0,2);
```

Is ``ds`` has a sign and is not an absolute value. It is always negative, if the source is "behind" the array (general case for a non-focussed point source). Hence ``dref - ds`` makes sense and is correct. If you want to change it to ```dref + ds`` then you also have to change definition of ``ds``. 

The correction factor for focussed sources is not the same as for point sources. The minus sign is correct here imho.

hagenw:
As @fietew pointed out, these changes were wrong.

fs446:
This gain factor occurs for virtual point source and the parallel reference line. The mismatch/confusion is due to different definitions and conventions of the used coordinates. Both are correct under the following assumption/setups:

> \sqrt(\frac{\xref}{\xref-\xs})

is correct for a SSD on y-axis (or z-axis), a virtual spherical monopole in xy-plane (or xz-plane) for \xs<0 and a reference line at xref>0. Here, \xs and \xref are handled as coordinate system values w.r.t. the x dimension. See the stuff that led to eq. (2.140) in Schultz (2016). In Start (1997) is a sign typo w.r.t. this derivation, compare eq. (3.14) and Schultz (2016) eq. (2.139).

> \sqrt(\frac{\xref}{\xref+\xs})

is the correct version, when \xs and \xref are handled as the shortest possible distances! from the virtual point source to the SSD and the reference line to the SSD resp., cf. Fig. 2.3 in Schultz (2016). This factor is useful when the driving function should be given in in general vector notation, cf. eq. (2.131) in Schultz (2016). Eq. (3.16) in Start (1997) is essentially the same conclusion. This version is used in the SFS tollbox docu at the moment, right?

hagenw:
Yes, we use the shortest possible distance definition:
http://sfstoolbox.org/en/latest/#equation-D.wfs.ps.2.5D.refline

PR #126, opened 2017-01-05
Nearest-neighbour search for BRIR direction

VeraE:
I noticed that the nearest-neighbour search for the head direction of a BRIR boldly calculates a Euclidian distance out of angles in radians. I suggest the following correction where the angles are converted to cartesian coordinates first. Not my best day for naming of variables, so feel free to change them...

The second change prevents the warnings for BRIR head directions being out of range to display due to numerical inaccuracies only.


VeraE:
After consultation with @trettberg this change proved to be a bit shortsighted only fixing this one simple case. The function `findconvexcone()` has instead been modified by @trettberg to handle special cases such as spherical caps and 2D arrays of points and can now be used to select nearest neighbours as well as interpolation in 3D.
I will implement the new interpolation handling in the functions for HRIR and BRIR selection and we can discuss this new version. This is closely related to #76.

PR #128, opened 2017-01-30
fix wfs_fs bug and nfchoa gains

fietew:
According to the original publication ( see https://doi.org/10.1109/ASPAA.2011.6082325 ), Equation (10) and (11), some weights are missing in our NFCHOA time-domain implementation for the plane waves and point source. This pull request fixes this.

Execute ``test_imp_25d(true)`` to compare the spectra of 2.5D SFS techniques near the reference point with the ground truth (plane wave or point source).

There was also a bug in ``SFS_time_domain/driving_functions_imp/driving_function_imp_wfs_fs.m`` related to a wrong usage of the ``/``-operator.

### DIFF
--- a/SFS_time_domain/driving_function_imp_nfchoa.m
+++ b/SFS_time_domain/driving_function_imp_nfchoa.m
@@ -94,16 +94,16 @@ for n=1:order+1
     % Get the second-order sections for the different virtual sources
     if strcmp('pw',src)
         % === Plane wave =================================================
-        sos = driving_function_imp_nfchoa_pw(n-1,R,conf);
+        [sos,g] = driving_function_imp_nfchoa_pw(n-1,R,conf);
     elseif strcmp('ps',src)
         % === Point source ===============================================
-        sos = driving_function_imp_nfchoa_ps(n-1,R,r_src,conf);
+        [sos,g] = driving_function_imp_nfchoa_ps(n-1,R,r_src,conf);
     elseif strcmp('ls',src)
         % === Line source ================================================
-        sos = driving_function_imp_nfchoa_ls(n-1,R,r_src,conf);
+        [sos,g] = driving_function_imp_nfchoa_ls(n-1,R,r_src,conf);
     elseif strcmp('fs',src)
         % === Focussed source ============================================
-        sos = driving_function_imp_nfchoa_fs(n-1,R,r_src,conf);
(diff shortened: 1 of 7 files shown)


SFS_time_domain/driving_function_imp_nfchoa.m
```diff
@@ -89,7 +89,7 @@
 
 % Compute impulse responses of modal filters
 dm = [repmat(pulse,[order+1 1]) zeros(order+1,N-length(pulse))];
-for n=2:order+1
+for n=1:order+1
```
hagenw: I guess this change has some undesired consequences. Please try (I used octave):
```Matlab
>> conf = SFS_config;
>> x0 = secondary_source_positions(conf);
>> driving_function_imp_nfchoa(x0,[0 2.5 0],'ps',conf);
error: driving_function_imp_nfchoa_ps: A(I,J): row index out of bounds; value 1 out of bound 0
error: called from
    driving_function_imp_nfchoa_ps at line 100 column 20
    driving_function_imp_nfchoa at line 100 column 13
```

If you switch back to master and run the same commands you won't get an error.

SFS_time_domain/driving_function_imp_nfchoa.m
```diff
@@ -151,7 +151,7 @@
 end
 d = circshift(d,[-order,0]);
 d = ifft(transpose(d),[],2);
-d = 1/pi/R*nls*real(d);
+d = 1/(2*pi)/R*nls*real(d);
```
hagenw: Maybe this is easier to read if you change it to
```Matlab
d = 1/(2*pi*R)*nls*real(d);
```

SFS_time_domain/driving_function_imp_nfchoa.m
```diff
@@ -166,5 +166,5 @@
         end
         d(:,n) = transpose(dtemp);
     end
-    d = 1/pi/R*real(d);
+    d = 1/(2*pi)/R*real(d);
```
hagenw: Here the same, maybe better to write
```Matlab
d = 1/(2*pi*R)*real(d);
```

fietew:
@hagenw: Could you please try again?

hagenw:
For me, everything is now working under Octave.
And I can also see the differences in the plots of `test_imp_25d(true)` if I enable or disable the gain corrections.
I'm going to merge this now.

PR #130, opened 2017-02-06
New interpolation handling

VeraE:
@trettberg and me continued the work on enabling interpolation in 2D and propose this as the new way interpolation is handled. This is related to #76.

It is now possible to interpolate HRTFs and BRIRs between three points on sphere. The three points and their weights are selected by Delaunay triangulation by the function `findconvexcone()`. The function is also able to interpolate between points located on a spherical cap or on a cirlce/arc (using two points only then). Test cases are demonstrated in the validation script `test_interpolation_point_selection()`.
The function `findnearestneighbour()` has been modifed to return weights and indices in the same way as `findconvexcone()`. This enables a concise implementation of the interpolation procedure itself. Only the function `interpolate_ir()` is used, `interpolation()` is unnecessary now. 
A new config struct parameter `conf.ir.interpolationpointselection` has been added to choose between the point selection methods.

We are thinking about changing the name of `findconvexcone()` since it is not perfectly suitable anymore. Maybe `findsurroundingsimplex()` or `findenclosingsimplex()`? Any suggestions are welcome.

What do you think of these changes for interpolation handling?


hagenw:
Cool, I will start adding a few comments to the single changes.

SFS_ir/sofa_get_head_orientations.m
```diff
@@ -1,20 +1,21 @@
...
+%       sofa   - impulse response data set (SOFA struct/file)
+%       idx    - index of measurement for which head orientation should be
+%                returned (default: return all head orientations)
 %
 %   Output parameters:
-%       phi     - head orientations in the horizontal plane / rad
-%       theta   - head orientations in the median plane / rad
+%       phi    - head orientations in the horizontal plane / rad
+%       theta  - head orientations in the median plane / rad
+%       r      - length of head orientation vector / m
```
hagenw: "length of head orientation vector" might be misleading as you would expect it to be something like `length([phi theta])`. Maybe you change this to `r      - head orientations radius / m`?

SFS_ir/interpolate_ir.m
```diff
@@ -1,33 +1,36 @@
...
 %
-%   INTERPOLATE_IR(ir,x0,xs,conf) interpolates the two to three given impulse
-%   responses from ir with their corresponding angles x0 for the given angles
-%   xs and returns an interpolated impulse response.
-%	For the 1D case, the interpolation method differs depending on the setting
-%	of conf.ir.interpolationmethod:
+%   INTERPOLATE_IR(ir,x0,xs,conf) interpolates the given impulse responses by
+%   applying the given weights and returns the interpolated impulse response as
+%   well as its corresponding position. Only impulse responses with weights >= prec
+%   will be used.
```
hagenw: It is not obvious to the user what `prec` might be as it is not something that can be configured, but is hardcoded in the function itself. This is fine, but maybe you should replace `prec` here by its actual value `0.001`?

SFS_ir/get_ir.m
```diff
@@ -124,28 +140,53 @@
...
+    [x0(:,1),x0(:,2),x0(:,3)] = cart2sph(x0(:,1),x0(:,2),x0(:,3));
+    [xs(1),xs(2),xs(3)] = cart2sph(xs(1),xs(2),xs(3));
     ir = ir_correct_distance(ir,x0(idx,3),xs(3),conf);
-    [ir,x0] = interpolate_ir(ir,neighbours,xs(1:2)',conf);
-    [x0(1),x0(2),x0(3)] = sph2cart(x0(1),x0(2),xs(3));
+    [x0(:,1),x0(:,2),x0(:,3)] = sph2cart(x0(:,1),x0(:,2),x0(:,3));
+    % Select or interpolate to desired impulse response
+    [ir,weights_selected,x0_selected] = interpolate_ir(ir,weights,x0(idx,:),conf);
+    % Calculate position of returned impulse response
+    x0_new = weights_selected'*x0_selected;
```
hagenw: You changed the return value of `interpolate_ir()` from `x0_new` to `weights_selected` and `x0_selected`. Now you have to calculate `x0_new` manually here.
Otherwise the new return values are not really used inside `get_ir()`, also not at the other position where `interpolate_ir()` is used. What was the motivation to change the output parameter of `interpolate_ir()`? Can we maybe return to the old behavior?

SFS_ir/sofa_get_head_orientations.m
VeraE: You are right, that's midleading. I changed it.

SFS_ir/interpolate_ir.m
VeraE: I changed that, too.

SFS_ir/get_ir.m
VeraE: Hm, I know that part is not very pretty. The problem is, that with the new calculation of weights for points used for interpolation, a linear combination of the weighted points does not yield the desired new point but it lies on the direct connection between two points or on the plane spanned by three points. That is, the radius is not what was intended. Only the direction is relevant for the interpolation weights. Adjusting of the amplitude due to a different source to listener distance is done by `ir_correct_distance()`.
And as I thought the calculation of `x0_new` with fixing of the radius should not be inside of `interpolate_ir()`, I calculated it this way.

hagenw: You mean the calculated radius value of `x0_new` is false and that is the reason why you replace it with `x0_new(3) = xs(3);` in line 179, correct?

VeraE: Yes!

SFS_ir/interpolate_ir.m
```diff
@@ -88,71 +91,41 @@
...
+        idx_half = floor(size(TF,3)/2)+1;
+        magnitude_new = sum(bsxfun(@times,magnitude(:,:,1:4:idx_half),weights),1);
+        phase_new = sum(bsxfun(@times,phase(:,:,1:4:idx_half),weights),1);
+        % Calculate interpolated impulse response from new magnitude and phase
+        ir_new = ifft(magnitude_new.*exp(1i*phase_new),size(ir,3),3,'symmetric');
+    otherwise
+        error('%s: %s is an unknown interpolation method.', ...
+            upper(mfilename),interpolationmethod);
     end
+elseif ~useinterpolation || length(weights)==1
```
hagenw: This should be the first `if`-clause otherwise the case `length(weights)==1` is only be considered if `~useinterpolation`.

SFS_ir/get_ir.m
hagenw: OK.
What will happen if the two IRs used for interpolation don't have the same radius?
Then the `simple` interpolation method makes no sense, but will still return something. What is with the `freqdomain` method, is this working with different radii?
Otherwise we should maybe return an error if the radii are not the same. And I guess then we could also move the corrected calculation of `x0_new` to `interpolate_ir()`.

fietew:
I would propose to use ``.'`` or ``transpose()`` instead of ``'`` to transpose matrices. ``'``yields in the conjugate transposed matrix, which for real values is no big deal, but using ``.'``  or ``transpose()`` consistently could prevent bugs.

SFS_ir/get_ir.m
VeraE: The `freqdomain` interpolation method can handle different radii but it cannot be used this way in the SFS toolbox because we are using `ir_correct_distance()` to handle adjustment of distances.

`findconvexcone()` is returning an error if the points used for interpolation do not lie on a sphere. `findnearestneighbour()` ignores the radius for calculation of the 1D weights but it is stated `SFS_config()` that this point selection method is only to be used for points on a circle.

`interpolate_ir()` would need `xs(3)` to calculate `x0_new`. To me, this would make the function unclearer.

SFS_ir/interpolate_ir.m
VeraE: Ah, you are right, I changed that.

VeraE:
Is the use of `.'` or `'` already consistently used in the whole SFS toolbox? Because there will be no problem in the new code with it as it is at the moment.

SFS_ir/get_ir.m
hagenw: In the former version of `get_ir()` different radii were ignored by simply passing `x0(:,1:2)` in spherical coordinates to the point search algorithms. This was not correct as it ignored the 360° wrap problem of the angles, but I was wondering if something similar might still be possible. Can we maybe do a search in the phi,theta space without the wrapping problems?

hagenw:
I'm still not so comfortable with the new version of `get_ir()` there are lots of `cart2sph()` and `sph2cart()` entries which looks not really elegant.
In addition the code for the part on `% Get nearest neighbour point or points for interpolation` is repeated for `SimpleFreeFieldHRIR` and `MultiSpeakerBRIR`.
The last point can be solved by creating an additional function. For the first one, I'm not sure what might be the easiest solution. For example, the part starting at line 146 that tries to incorporate the head orientation, is this only possible in spherical coordinates or can it maybe directly be included in line 142?

SFS_ir/get_ir.m
VeraE: I'm not sure if I understood what you meant: Do you think about passing the points `x0` and `xs` to the point selection functions in spherical coordinates so the conversion to cartesian coordinates is done there?

fietew:
@VeraE: Neither is, I think. Maybe this pull request is a good opportunity to agree on a consistent use of ".'" for this and future pull requests. Off course, revising the whole existing code is not worth the effort, imho.

fietew:
> I'm still not so comfortable with the new version of get_ir() there are lots of cart2sph() and sph2cart() entries which looks not really elegant.

I think, the default conversion strategy starting in https://github.com/sfstoolbox/sfs-matlab/blob/a7ac6bf764de0136782880d765e37dc1a86604ef/SFS_ir/get_ir.m#L116 (from spherical to cartesian coordinates) might not be optimal. One could save many conversions, if the coordinates are converted to spherical coordinates by default.

VeraE:
1. I can create a new function for the repeated part of the point selection. I think I better do this after we agreed on the structure of the rest of the function.

2. I really thought a while about the conversions from spherical to cartesian coordinates and back. I don't like it that much either but I could not come up with a better way that is at the same time easily to understand if you go through the code line by line. This too applies to the example of the incorporation of head orientation starting at line 146. Are you concerned about it because it could be more efficient or because you think it's confusing?
And I don't think that it is better to start with spherical coordinates in `get_ir()` as cartesian coordinates are needed in the beginning at line 142. @fietew Or maybe you have got an easy solution to combine lines 142-150 in one step in spherical coordinates? That would be great!

3. I'll go through the code in this pull request and make sure `'` and `.'` are used consistently.

hagenw:
1. @VeraE I think you can start already writing the function as I don't think we will change the whole structure of `get_ir()`. Maybe it is a good idea to write the function as a subfunction at the end of `get_ir()`?

2. My main concern with all these coordinate system switches is that we have a very high numbers of line numbers and the code is becoming harder to understand. The performance I have not tested yet, maybe we should also check this.
I was never good at coordinate system transformation, and I'm also wondering if line 142-150 could be done in one system.

VeraE:
1. I just moved the point selection part to a subfunction.

2. Maybe there exists a nice transformation to avoid some coordinate system switches, but same as you, I am not very good at that which is of course one reason why I had to write it this way. The other reason is that I prefer some more simple lines of code over a short elegant version that needs additional knowledge to understand as long as it is not a matter of efficiency and I do not consider the coordinate system switches as relevant to efficiency.
I tested the calculation of an SSR ready data set (161 head angles in horizontal plane, WFS array with 16 loudspeakers) with nearest neighbours interpolation in this here new version against the master branch. The new version was faster by about 30% which is quite nice I think.

hagenw:
1. Cool, looks good.

2. For me it would be fine to merge the current version and create an issue reminding us, that there is room for improvements regarding coordinate transformations.

hagenw:
I have another comment regarding the return value `x0_new`. It might be a little bit confusing that we return only the loudspeaker position and not the head orientation. And for SimpleFreeFieldHRIR we are mixing loudspeaker position and head orientation, but of course we are not doing it for MultiSpeakerBRIR.

I have further checked all other functions of the Toolbox and found no single usage of the return value `x0_new`. So maybe we can make `get_ir()` a little bit easier by just removing that parameter altogether.
@VeraE what do you think?

VeraE:
Creating an issue as reminder for improvement of the coordinate transforms is fine with me, too.

Regarding `x0_new`: I considered this only as an informative value. It was there before so I kept it the same way. I just had to change the way it was calculated. It's fine with me to delete it, though. This would indeed make the function clearer to read. Do you just delete it?

hagenw:
Yes, I will do it.

SFS_ir/interpolate_ir.m
```diff
@@ -88,71 +91,41 @@
...
+        % Magnitude and phase will be interpolated separately
+        magnitude = abs(TF);
+        phase = unwrap(angle(TF),[],3);
+        % Calculate interpolation only for the first half of the spectrum
+        % and only for original bins
+        idx_half = floor(size(TF,3)/2)+1;
+        magnitude_new = sum(bsxfun(@times,magnitude(:,:,1:4:idx_half),weights),1);
+        phase_new = sum(bsxfun(@times,phase(:,:,1:4:idx_half),weights),1);
+        % Calculate interpolated impulse response from new magnitude and phase
+        ir_new = ifft(magnitude_new.*exp(1i*phase_new),size(ir,3),3,'symmetric');
```
hagenw: The option `'symmetric'` is not available for `ifft()` under Octave. Is there a workaround?

hagenw: The option `'symmetric'` is not available for `ifft()` under Octave. Is there a workaround?

VeraE: Really? This option was already in our code before. The spectrum has to be mirrored manually which means some additional lines of code. Shall I do that?

hagenw: I found it also in another function of the Toolbox.
I will first have a look if there is a workaround directly under Octave, so you don't have to do it.

chris-hld: `signal_from_specrum.m` takes care of the correct mirroring as well as potential round-off errors. Could that help?

PR #131, opened 2017-02-15
use persistent variables to speed up resampling filter

fietew:
MATLAB's profiler told me, that the design of the resampling filter which is done in ``firpm`` for each ``delayline`` call takes at lot of time. This fix caches the filter coefficients in a persistent variable.


fietew:
The function uses normalised frequencies as an input (so between 0.0 and 1.0, while 1.0 corresponds to fs/2). ``Wpass`` and ``Wstop`` are taken from the ``fdatool`` (filter design tool) of MATLAB.

PR #134, opened 2017-03-02
DOC: new section on system impulse responses and switch to opengl

hagenw:
This is a follow up on #118 and picks only the really needed commits.


hagenw:
@chris-hld could you please run `generate_plots` and commit all the figures here, thanks.

hagenw:
Thanks.

PR #135, opened 2017-03-23
Time-Domain Implementation of LocalWFS using Spatial Bandwidth-Limitation

fietew:
Yet another method for Local Sound Field Synthesis. Here, the plane wave decomposition of a modally truncated sound field is reproduced via conventional Wave Field Synthesis. Due to the modal truncation, like in NFCHOA, spatial aliasing is avoided.

TODOs:

* [x] documentation in ``SFS_config`` **see 964447e**
* [x] ``conf.t0`` should be taken into account **add error message, see 2659c2d**
* [x] the naming scheme for the already existing Local WFS should be changed. I would suggest something like ``_localwfs_vss`` as this approach uses virtual secondary sources. **Added new function calls and warnings, that former function calls are deprecated, see 6bc0908**
* [x] more testing functions?
* [x] add modal weighting function **see 428e8c5**


hagenw:
I had a first look at it and corrected formal stuff. Those points are still pending and it would be nice if you could have a look at them:

* In `inverse_cht()` the header of the function is missing.
* In `driving_function_imp_localwfs_sbl_ps()` the function header is missing.
* In `driving_function_imp_localwfs_sbl_ps()` there  is a TODO (ensure that the time-reserved filter is not truncated)
* In `SFS_time_domain/driving_functions_imp/driving_function_imp_localwfs_sbl_pw.m()` the function header is missing.

hagenw:
By running `test_imp_25d()` I got the following error:
```
error: DRIVING_FUNCTION_IMP_LOCALWFS_SBL: conf.t0 (system) other than "source" is not supported
error: called from
    driving_function_imp_localwfs_sbl at line 67 column 5
    test_imp_25d at line 117 column 11
```

I added `conf.t0 = 'source';` to avoid this.

hagenw:
In `test_imp_25d()`, if I change `conf.delayline.resampling = 'pm';` to `conf.delayline.resampling = 'matlab';` I'm getting completely different results. Is this the same for you? The results with `pm` look a lot better.

hagenw:
`test_imp_25d()` is still not running to the end for me and I get
```
error: 'linkwitz_riley' undefined near line 102 column 20
error: called from
    circexp_imp_ps at line 102 column 18
    driving_function_imp_localwfs_sbl_ps at line 76 column 19
    driving_function_imp_localwfs_sbl at line 73 column 21
    test_imp_25d at line 118 column 11
```
Is `linkwitz_riley()` part of Matlab or do you have forgotten to add it?

hagenw:
Thanks, now I'm getting the following error with `test_imp_25d(1)`:
```
error: bilinear: must have at least as many poles as zeros in s-plane
error: called from
    bilinear at line 91 column 5
    circexp_imp_ps at line 114 column 15
    driving_function_imp_localwfs_sbl_ps at line 76 column 19
    driving_function_imp_localwfs_sbl at line 73 column 21
    test_imp_25d at line 118 column 11
```

fietew:
I can't reproduce this error under MATLAB. Is it octave related?

hagenw:
Yes, I remember. Could you have a look at [`bilinear_transform()`](https://github.com/sfstoolbox/sfs-matlab/blob/localwfs-sbl/SFS_general/bilinear_transform.m) and use that one instead. There it is internally managed to distinguish between Matlab and Octave as the needed function call to `bilinear()` is different.

fietew:
This straightforward, as ``bilinear_transform()`` does required second order sections and does not return zero-pole-gain representations. I will try to handle the octave stuff directly.

fietew:
The error occurs, for ``n=0`` inside the loop of ``circexp_imp_ps``. It seems that octave does not handle the case, where there is no zero and no pole, correctly.

fietew:
@hagenw: I'm still getting an error under octave which is related to my (maybe broken) octave installation. Could you give it a try?

hagenw:
The problem with calling `bilinear()` seems to be fixed now, but I'm getting a new error:
```
error: Invalid call to sosfilt.  Correct usage is:

 -- Loadable Function: Y = sosfilt (SOS, X)
error: called from
    print_usage at line 90 column 5
    circexp_imp_ps at line 128 column 15
    driving_function_imp_localwfs_sbl_ps at line 91 column 19
    driving_function_imp_localwfs_sbl at line 77 column 21
    test_imp_25d at line 118 column 11
```

And one question regarding your fix: under Octave you are now simply ignoring `m=0`. What are the consequences? Could you ignore it also under Matlab?

fietew:
No, I'am not ignoring it. For ``n=0``, ``zh`` and ``ph`` are empty before and after the bilinear transform. If there are no zeros/poles in the s-domain, there will be no zeros/poles in the z-domain.

fietew:
I fixed the call of ``sosfilt`` for octave, but the results for the point source in octave and MATLAB do not coincide.

[matlab.pdf](https://github.com/sfstoolbox/sfs-matlab/files/958254/matlab.pdf)
[octave.pdf](https://github.com/sfstoolbox/sfs-matlab/files/958249/octave.pdf)

fietew:
``butter()`` yields row vectors in octave instead of column vectors in MATLAB.

hagenw:
Cool, it is working now.

hagenw:
You asked if we should provide more test functions. Maybe it is a good idea to add a test function for `circexp_imp/` and `pwd_imp/` as those are more general and not only connected to LSFS?

fietew:
added new test function and resolved conflicts with master

hagenw:
The new test function is working fine for me. Maybe you could add a title to every plot like we did it in `test_driving_functions()`. Then it will be easier to interpret the single plots.

PR #137, opened 2017-04-05
Enable shortcut for exact point

VeraE:
I had to make a small change so that the interpolation routine is actually skipped when the desired point is contained in the selected points and no interpolation is necessary, that is one weight is 1 and the others are zero.


hagenw:
There still seems to be a bug to me. In line 102 we define `weights_new` that you now introduced correctly to line 106 as well. But on lines 111, 123 and 124 we are still using `weights` which seems to make no sense to me, or am i wrong?
Maybe we should get rid of all the `_new` variables and use `weights = weights(weights>=prec);`, ` x0 = x0(weights>=prec,:);` and so on instead?

VeraE:
You are right, there was still a bug, I fixed it now and got rid of the `_new` variables just as you suggested. We are not returning those to `get_ir()` anyway.

hagenw:
Thanks, I did a further simplification of the `if` clause and removed more `_new`. Could you please have a look if this is still fine with you and test if it is still working correctly.

VeraE:
Thanks, looks good. Now that we are not returning the used positions `x0`, I deleted it.

PR #138, opened 2017-04-27
fix incorporation of tapering weights for LWFS

fietew:
As ``x0s(:,7)`` also contains the integration weight of the secondary source, the influence of the tapering has to be computed by normalising the result to the original integration weight.

### DIFF
--- a/SFS_monochromatic/driving_functions_mono/driving_function_mono_wfs_vss.m
+++ b/SFS_monochromatic/driving_functions_mono/driving_function_mono_wfs_vss.m
@@ -83,11 +83,6 @@ else
     error('%s: %s is not a known source type.',upper(mfilename),dimension);
 end
 
-% Adjust weights of secondary sources in order to use the tapering
-% correctly. Integration weights for the secondary sources will be applied
-% later, when the sound field is computed
-x0(:,7) = 1;
-
 % Get driving signals for real secondary sources
 %
 % See Spors (2010), fig. 2 & eq. (12)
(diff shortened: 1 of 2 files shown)


hagenw:
Has this also to be considered in `SFS_monochromatic/driving_function_mono_localwfs.m`?

fietew:
It is already considered by https://github.com/sfstoolbox/sfs-matlab/blob/2bbd541d3aa1783a7b5273339b316e3638096d9f/SFS_monochromatic/driving_functions_mono/driving_function_mono_wfs_vss.m#L89.

hagenw:
If you would like to use only the tapering weights and exclude the integration weights you might also considering to use `tapering_window()` instead of `secondary_source_tapering()`. This is actual used in the latter, [have a look](https://github.com/sfstoolbox/sfs-matlab/blob/master/SFS_general/secondary_source_tapering.m#L58).

For the time domain the code would then look like this:
```Matlab
wtap = tapering_window(x0s,conf);
```
and you can remove the `x0tmp = secondary_source_tapering(x0s,conf);` line.

For the frequencz domain you could change it to
```Matlab
wtap = secondary_source_tapering(xtmp,conf);
Dmatrix(xdx,idx) = driving_function_mono_wfs(xtmp,xv(idx,1:3),'fs',f,conf) .* wtap;
```
and remove the `x0(:,7) = 1;` line. In my opinion the result would be easier to understand.

fietew:
Ah, cool. I will change both functions.

Issue #136, opened 2017-03-31
Unified 2.5D WFS Framework


fs446:
We might add the driving functions of plane wave, point/line source of the unified 2.5D WFS framework for different referencing schemes / positions of amplitude correct synthesis (PCS) into the toolbox based on the findings 
[https://doi.org/10.1109/TASLP.2017.2689245](https://doi.org/10.1109/TASLP.2017.2689245)
, see also 
[142nd Audio Eng. Soc. Conv, Berlin, #9722](http://www.aes.org/events/142/papers/?ID=5360)
for general vector notation of these driving functions. Then, 'only' the referencing function must be adapted to work in general vector notation for desired reference curves.

hagenw:
Nice idea.
@fs446 or @gfirtha do you have time to create a small first pull request with an example implementation, or should I have a try at it?

fs446:
Good. We’ll initially provide some stuff, like 

get_unified_25D_WFS() general driving function
get_dir_derivatives() for point/line/plane 
set_ref_function()  for different feasible referencing schemes and SSD contours, that’s the most critical part to solve for arbitrary setups
get_PCS() positions of amplitude correct synthesis for check and debug, not known before in detail

Hope this fits into the actual framework. Once this is realized, all currently used 2.5D WFS driving functions just need a customized set_ref_function()-case to work within this framework. That was in fact the idea ;-)

fs446:
init dev in branch 'unified_25d_wfs'

fs446:
[D, xPCS] = driving_function_mono_unified_25d_wfs(x0,xs,dx0,src,f,conf) is working as expected with the current state dab14a1cb411058083f70c9a60e32347aff473d0
Fig. 3,4,9,10 from https://doi.org/10.1109/TASLP.2017.2689245 created with this one->ok

Now we need a meaningful implementation for defining a referencing function somewhere outside the calculation of the sound field with sound_field_mono_wfs(X,Y,Z,xs,src,f,conf). A further input parameter could do this. Any ideas, how we should proceed to be most consistent with the current toolbox handling.

fs446:
added a test script in 916fc76cb3d3d9f3afbeb64b63671b28bf71dbc6 that checks the driving functions.

hagenw:
Cool, thanks for your effort. I will have a look at this next week.

hagenw:
[...]

PR #143, opened 2017-05-05
Unified 25d WFS referencing scheme

hagenw:
For the start of the discussion have a look at #136.

### DIFF
--- a/SFS_monochromatic/driving_function_mono_unified_25d_wfs.m
+++ b/SFS_monochromatic/driving_function_mono_unified_25d_wfs.m
@@ -1,24 +1,31 @@
-function D = driving_function_mono_unified_25d_wfs(x0,xs,src,f,conf)
+function [D, xPCS] = driving_function_mono_unified_25d_wfs(x0,xs,dx0,src,f,conf)
 %DRIVING_FUNCTION_MONO_UNIFIED_25D_WFS returns the driving signal D for WFS
 %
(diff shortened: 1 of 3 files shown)


fietew:
As far as I could follow the discussion in #136, the integration of the code into the existing framework is still an issue. My suggestion to integrate this unified scheme a little bit more into the existing code would be the following (asterisk for ``driving_function_mono_`` or ``driving_function_imp_``):
* extend ``*_wfs_<src>``, etc. about the optional input parameter ``dx0``
* in ``*_wfs_<src>`` for ``conf.dimension`` = ``'2.5D'``: 
    * compute pre-filter
    * compute the directional gradient of the desired sound field (as done in ``driving_function_mono_unified_25d_wfs``)
    * If ``dx0`` is given, ignore ``conf.driving_function``. If no ``dx0`` is given, compute ``dx0`` based on ``conf.driving_function``.
    * compute ``xPCS`` and driving function

The drawback would be, that some driving functions, e.g. for point source and focused source, might not fit into this framework.

hagenw:
I'm also in favour to integrate this in the existing framework.
This pull request will not be merged into master, but we should integrate the changes in another branch I would suggest. I created this pull request mainly in order to easily review the changes proposed by @fs446.
Unfortunately, I'm still a little bit short with time at the moment, but in a few weeks I should also be able to work on this issue. I would propose that we maybe first prepare a new release of the toolbox including all the LSFS changes and after that start integrating this one. Does this sound ok?

PR #148, opened 2017-05-12
add max-rE and modified Tukey window

fietew:
max-rE weighting has been widely used in the (NFC-HO)Ambisonics community. I added it for comparison. The Tukey window is a second adjustable window, which might have some benefits compared to the Kaiser windows as its inner part (m < alpha*M) is always one.


fietew:
@fs446: Many window functions (number of samples denoted as N) are defined such that its first (n=0) and last (n=N-1) value are zero, e.g. many windows with a cosine shaped fade out involved. Is it in general sensible to modify the cosine fade out such that it yields non-zero values for n=0 and n=N-1, but would yield zero for n=-1 and n=N?

fs446:
That’s indeed an issue, but not often suffiently discussed in literature. So, it’s good to open this up here. I remember that Ifeachor, Digital Signal Processing: A Practical Approach has short comments on this. And the script of Michael Möser’s course on signal processing is brilliant regarding this issue https://www.akustik.tu-berlin.de/fileadmin/fg23/Lehre/Digitale_Signalverarbeitung_gesamt.pdf, from ch. 2.4 signal model.
See also his habilitation script 'Analyse und Synthese akustischer Spektren' at Springer 1988. Unfortunately all in german.

The essence is:
How many of the available zeros of the weighting polynomial are to be used and where they should be located within the unit circle?

> Is it in general sensible to modify the cosine fade out such that it yields non-zero values for n=0 and n=N-1, but would yield zero for n=-1 and n=N?
I would precisely do this for FIR filters or gain shading windows for acoustic array (which is an FIR filter :-) ), since it is useless to have weights/taps with zero gain. Then you'd use all potential zeros for shaping the spectrum.

For DFT-windowing I’d rather use a window that starts with zero at n=0 and gets to zero at n=N, thus having non-zero value at n=N-1. That’s the ‘periodic’ flag in Matlab.

hagenw:
As max-rE is indeed widely used we might also considering adding an example to the documentation?

fietew:
I tried to add a plot to ``generate_plots.m``, but getting the error:
```
Undefined function or variable 'xticks'.

Error in generate_plots (line 24)
xticks(-2:1:2); yticks(-2:1:2);
```
I don't think, that this is related to the changes.

hagenw:
That's true, the `xticks()` was introduced only in MatlabR2016b. So maybe we should replace it with a version that also works on older versions. Do you know how to achieve the same in older versions?

fietew:
I hope, that ``d9a4014`` does the trick.

hagenw:
Yes, this works for me. But know I'm getting the following error:
```Matlab
> In sound_field_mono_localwfs (line 36)
  In generate_plots (line 175) 
Variable output array "varargout" must be a cell array.

Error in sound_field_mono_localwfs

Error in generate_plots (line 175)
sound_field_mono_localwfs([-1 1],[-1 1],0,[1.0 -1.0 0],'pw',7000,conf);
```

fietew:
I don't know, where to include such example because the modal window function has no effect on the monochromatic NFC-HOA driving functions.

hagenw:
I would say we have two options. Wait until we have a monochromatic implementation of it or go with an example in the time domain. As far as I understand it, this example should use a low number of loudspeakers as those are the systems that are normally used with max-rE, correct? And if we do an example with and without weighting I guess you should be able to see that with weighting not all loudspeakers are active any longer.

fietew:
I added a new subsection for the modal windows in NFC-HOA. This could be further extended, when adding documentation about LWFS-SBL. Latter should however not be handled in this pull request.

Issue #117, opened 2016-12-07
3D time-domain NFCHOA driving function


vancymoon:
When I try to use the  `ir_nfchoa` function

```MATLAB
ir = ir_nfchoa([0 0 0],0,[3 3 -3],'ps',hrtf,conf);
```
Errors occur:

```MATLAB
Warning: SOFA 0.6 upgraded to 1.0 
> In SOFAupgradeConventions (line 145)
  In SOFAload (line 186)
  In getting_start (line 9) 
Error using to_be_implemented (line 55)
This functionality has to be implemented yet!

Error in driving_function_imp_nfchoa_ps (line 116)
        to_be_implemented;

Error in driving_function_imp_nfchoa (line 97)
        sos = driving_function_imp_nfchoa_ps(n-1,R,r_src,conf);

Error in ir_nfchoa (line 77)
d = driving_function_imp_nfchoa(x0,xs,src,conf);

Error in getting_start (line 12)
ir = ir_nfchoa([0 0 0],0,[3 3 -3],'ps',hrtf,conf);
```
How can I solve this problem?

hagenw:
Hi,

as I deduce from your error message ([line 116 in `driving_function_imp_nfchoa_ps`](https://github.com/sfstoolbox/sfs-matlab/blob/master/SFS_time_domain/driving_functions_imp/driving_function_imp_nfchoa_ps.m#L116)) you are trying to simulate 3D NFC-HOA, so I guess you set `conf.dimension = '3D'`.

The error occurs as we have only implemented the [2.5D variant](https://github.com/sfstoolbox/sfs-matlab/blob/master/SFS_time_domain/driving_functions_imp/driving_function_imp_nfchoa_ps.m#L99) of a point source for NFC-HOA in 2.5D in the time domain. Unfortunately, 2D and 3D are not yet implemented in the time domain, so you cannot create binaural simulations for those setups and NFC-HOA, but only for a 2.5D setup (which means using a circular loudspeaker array consisting of point sources).

@fietew: are we able to implement the 3D variant or will it be complicated?

fietew:
[...]

PR #149, opened 2017-05-12
3D time-domain NFCHOA driving function

fietew:
Follow up to #117: Supports plane waves and point sources

use ``test_driving_functions_imp_with_delay`` to test


fietew:
There seems to be a bug ``secondary_source_positions``: The additional weighting with ``cos(theta)`` for spherical arrays is wrong imho. It leads to different results if the source is not in the xy-plane. I added two additional examples to ``test_driving_functions_imp_with_delay``.

hagenw:
You removed the additional weighting now completely. Do we not need some kind of weighting for integration on a sphere?

fietew:
The integration weights are defined by the quadrature scheme, i.e. the spherical grid. The different grid should already yield the correct integration weights. The ``cos(theta)`` weighting is only correct for Gauss grids, imho. Try ``[, w] = legpts(100), plot(w)``

hagenw:
OK, I see. We have a Gaussian grid in the toolbox, are the weights correctly given there?
See https://github.com/sfstoolbox/sfs-matlab/blob/master/SFS_general/get_spherical_grid.m#L108

fietew:
I tried this:
```
conf = SFS_config;

conf.secondary_sources.geometry = 'sphere';
conf.secondary_sources.size = 2.0;  % unit sphere

conf.secondary_sources.grid = 'gauss';
conf.secondary_sources.number = 200;
sum(x0(:,7))/(4*pi)

conf.secondary_sources.grid = 'fabian';
conf.secondary_sources.number = 11345;
sum(x0(:,7))/(4*pi)

conf.secondary_sources.grid = 'equally_spaced_points';
conf.secondary_sources.number = 900;
sum(x0(:,7))/(4*pi)
```
The FABIAN grid has some issues. However, it is not a full sphere grid.

hagenw:
The problems with the FABIAN grid existed before as well. If I execute your code with the master I'm getting:
```
0.7858
96.7188
0.7854
```
and now
```
1.0000
141.2893
1.0000
```
So, your fix for the secondary source integrations weights seem to be correct.

Regarding the integration weights for FABIAN, I guess there where given together with the data set, but I'm not sure. In the [official release of the data set](http://dx.doi.org/10.14279/depositonce-5718) the state that the weights where calculated by rectangles around the grid points.

SFS_time_domain/driving_function_imp_nfchoa.m
```diff
@@ -129,24 +132,38 @@
...
+    %                1     \
+    % D(phi0,w) = -------  /__     Hm(w) e^(im(phi0-phi_src))
+    %             2 pi r0  m=-M..M
+    %
+    % see Spors et al. (2011), eq.(4) and (5)
+    %--------------------------------------------------------------------------
+    
+    % apply modal weighting
+    wm = modal_weighting(order,conf);  % modal weighting function
+    Hm = bsxfun(@times,Hm,wm);
```
hagenw: Should we move this before the if clause? I guess the modal weighting should be applied in the 3D case as well.

SFS_time_domain/driving_function_imp_nfchoa.m
```diff
@@ -129,24 +132,38 @@
...
+    % apply phase shift due to source angle
+    m = 0:order;
+    dm = bsxfun(@times,Hm,exp(-1i*m*phi_src));
+    
+    % Spatial IFFT
+    dm = [conj(dm(:,end:-1:2)), dm];  % append coefficients for negative m
+    d = inverse_cht(dm,nls);
+    d = 1/(2*pi*R)*real(d);
+
+elseif strcmp('3D',dimension)
```
hagenw: Could you add a comment with the actual function and a reference similar to the 2.5D case.

SFS_time_domain/driving_function_imp_nfchoa.m
fietew: The problem might be, that the max-re window in 3D is different from its 2D version. Shall we distinguish between 2D/3D directly in the ``modal_weighting``?

hagenw: Is this only the case for the max-re window?
I also see no apparent alternative. At the moment we simply use no weighting window in the 3D case which you would not expect as a user if you have set the conf setting to use one window.
If you would not distinguish between 2D/3D directly in `modal_weighting` how would you then handle it? Introduce another `modal_weighting_3D` function, or is it no longer called modal weighting in the 3D case.

fietew: Well it could be the case for other windows, but I'm not sure. It really depends on the optimisation criterion, which is fulfilled by a specific window function. I will include the 3D case inside ``modal_weighting``.

hagenw:
I tested it and it worked. In my opinion this is now ready to merge.

I'm tempted to clean up the commit history by some squashing, but as this includes a merge of the master and also some code from another pull request, the safest might be a merge?

hagenw:
I created a new branch `nfchoa_imp_3D_squash`, where I modified the commit history [a little bit](https://github.com/sfstoolbox/sfs-matlab/commits/nfchoa_imp_3D_squash). That branch I can easily rebase on master if this is fine with you.

fietew:
Please do not rebase the ``nfchoa_imp_3D_squash`` branch onto master, as this would destroy the connection between the changes and this pull request on github. Please modify the commit history of ``nfchoa_imp_3D`` in the same way as you have done it for ``nfchoa_imp_3D_squash`` and then rebase the former onto master.

hagenw:
That's true, thanks for the hint.

Issue #71, opened 2016-03-01
Calculating zeros of spherical Hankel function


hagenw:
This is a follow up on #57.

We calculate the zeros of the spherical Hankel function with `sphbesselh_zeros()`, which works only up to an order of 85 due to numerical problems.

For higher orders we have to find a solution, which will be discussed here.

hagenw:
I would like to start very simple with this discussion.

The problem we would like to solve is to find the zeros of the spherical Hankel function:
`h_n (z) = 0`                                                                                                                                      (1)
where `n` is the order.
The spherical Hankel function of first kind is defined as
`h^1_n (z) = j_n (z) + i y_n (z)`                                                                                                           (2),
where `j_n (z)` is the spherical Bessel function of first kind and `y_n (z)` is the spherical Bessel function of second kind.
The spherical Hankel function of second kind is defined as
`h^2_n (z) = j_n (z) - i y_n (z)`                                                                                                            (3).

Further the spherical Hankel functions are given as
`h^1_n (z) = \sqrt{\frac{2}{\pi z}} H^1_{n + \frac{1}{2}} (z)`                                                                 (4),
where `H^1_n (z)` is the Hankel function of `n`th order and first kind, and
`h^2_n (z) = \sqrt{\frac{2}{\pi z}} H^2_{n + \frac{1}{2}} (z)`                                                                 (5),
where `H^2_n (z)` is the Hankel function of `n`th order and second kind.

From those definitions the following questions arise:

1. Are the zeros from `h^1_n (z)` and `h^2_n (z)` identical? If not, for which zeros are we looking?
2. Is the definition in eq. (2) and (3) identical with the stating that `h^*_n (z)` has only zeros if `j_n (z)` and `y_n (z)` are both zero at `z`? If yes, then this would also answer question 1.
3. Eq. (4) and (5) seem to be a little bit unfortunate as they gave us a connection to `H^*_{n + \frac{1}{2}} (z)` instead of `H^*_n (z)` for which some zeros are available in the literature.¹ If we have zeros for `H^*_n (z)` is there a way to directly get the ones for `h^*_n (z)`?

¹ For zeros of `H^*_n (z)` see for example [Döring (1966)](http://www.ams.org/journals/mcom/1966-20-094/S0025-5718-1966-0192632-1/S0025-5718-1966-0192632-1.pdf).

fietew:
[...]

Issue #57, opened 2016-01-25
Second-order section structure filters for NFC-HOA driving function


narahahn:
The SOS coefficients in the Laplace domain is mapped to the SOS coefficients in the z-domain by using the bilinear transform. The latter are stored in two cell arrays, b and a. The filtering is then performed in a for-loop. I think we can do this in a simpler way:
* transform the zeros in the s-domain to the z-domain using the bilinear function
* compute the SOS coefficients using zp2sos (SOS is a matrix)
* filter the input (pulse) using sosfilt

hagenw:
As I have no experience with the SOS filtering stuff, could you maybe create a branch and implement an example of your solution?
The only thing we have to consider is to test it also in Octave. If I remember correctly the first implementation of the NFC-HOA stuff in the time-domain was only running under Matlab.

narahahn:
I just created a branch and updated the SOS implementation of 2.5D NFC-HOA driving functions.

Previously, the matlab function `zp2sos` was used with the pole-zeros in the Laplace domain. @fs446 pointed out that `zp2sos` is actually intended for z-domain representations (either pole-zeros or transfer function coefficients).(http://de.mathworks.com/help/signal/ref/zp2sos.html) Mathematically, the results are identical but there might be numerical problems.

Now the driving functions are computed as follows:

- The zeros (`z`) of the spherical Hankel function is computed. (s-domain)

- The poles and zeros of the driving function is obtained by scaling `z`. (s-domain)

- The poles and zeros are transformed into the z-domain either by bilinear transform (`default` driving function) or by matched z-transform (`matchedz` driving function). Now we are in z-domain!

- Second-order section coefficients (`sos`) are computed by `zp2sos`.

- Each numerator of the `sos` is normalized so that the gain at fs/2 is 0 dB.

- The impulse (`pulse`) is filtered with `sos`

```matlab
SFS_start;
conf = SFS_config;
conf.dimension = '2.5D';
conf.secondary_sources.geometry = 'circle';
conf.secondary_sources.size = 3;
conf.nfchoa.order = 16;
conf.plot.usedb = true;

x0 = secondary_source_positions(conf);
X = [-2 2];
Y = [-2 2];
Z = 0;
t = 200;

% virtual plane waves
xs = [0,-1,0];
src = 'pw';

conf.driving_functions = 'default';
sound_field_imp_nfchoa(X,Y,Z,xs,src,t,conf);
title('bilinear transform')
[...]

Issue #56, opened 2016-01-25
NFC-HOA orders for time-domain driving functions


narahahn:
Unlike for monochromatic cases, it seems that the Ambisonic order cannot be freely chosen for time-domain NFC-HOA driving function. It only works if the order is an integer multiple of the number of secondary sources, of vice versa. I started to implement for arbitrary integer orders. Would it be worth doing this?

hagenw:
That is another good point. It would be a very good idea as we would have a big problem otherwise.
A few comments:
1. I have fixed a small error, see e170daa9a2f3df8fdbcf218fdf151ffb0a08a3f5 and 78e2b5dad2e6ebcb2e3d9c22229adc86e728b285
2. As we have corrected the behavior of `nfchoa_order()` -- see 422b4f9ede7e4527f1d6591cccca8514e204c351 -- it will not return not a multiple of the secondary sources for band-limited NFC-HOA. For example:
```Matlab
>> conf = SFS_config;
>> sound_field_imp_nfchoa([-2 2],[-2 2],0,[0 -1 0],'pw',300,conf)
Error using driving_function_imp_nfchoa (line 157)
DRIVING_FUNCTION_IMP_NFCHOA: the given number of secondary sources (64) cannot be subsampled to 62 driving signals. Choose a NFC-HOA order that is a multiple of 30.

Error in sound_field_imp_nfchoa (line 94)
d = driving_function_imp_nfchoa(x0,xs,src,conf);
```
As you can see the numbers in the error message are now also broken, but the more important thing is that for band-limited NFC-HOA we cannot met the desired multiple condition. We have 64 secondary sources, which corresponds to an order of 31 (instead of 32 as it was before we corrected `nfchoa_order()`) which will not work with 64 secondary sources.

If you have an idea how to fix this, it would be great if you could push a new branch and create a pull request for this.

narahahn:
I created a branch `nfchoa_arbitraryorder`. Now, the time-domain driving functions can be computed for any orders.

```matlab
conf = SFS_config;
conf.secondary_sources.number = 64;
conf.driving_functions = 'default';
conf.plot.usenormalisation = false;
conf.plot.usedb = false;

x0 = secondary_source_positions(conf);
X = [-2 2];
Y = [-2 2];
Z = 0;
xs = [0,-1,0];
src = 'pw';
t = 200;
pulse = transpose(hann_window(5,5,10));

conf.nfchoa.order = [];
[...]

PR #152, opened 2017-06-23
More stable algorithm to calculate zeros of spherical hankel function

fietew:
Follow up on #71: 

Based on the [publication](http://www.aes.org/e-lib/browse.cfm?elib=18609) of @narahahn and @spors.

This is basically a port of a [scipy routine](https://github.com/scipy/scipy/blob/v0.19.1/scipy/signal/filter_design.py#L3666-L3702).

@narahahn: Could you try out some parametrisation and check, if it is working as expected?


fietew:
@hagenw: Is it okay to remove the other methods?

hagenw:
The old version of the function returned `z` and `p` both as column vectors. Your new implementation returns `z` as row vector and `p` as column vector. Is this desired or a bug?

If the new method works better than the old ones, I see no big deal in replacing them as all our methods we implemented so far were buggy.

hagenw:
Is the code ready for me to have a look at it?

It would be cool if you could add a test function for this, maybe doing something like the figures in #57 for example [those two](https://github.com/sfstoolbox/sfs-matlab/issues/57#issuecomment-186291311) or Figure 3 in @narahahn [publication](http://www.aes.org/e-lib/browse.cfm?elib=18609)

fietew:
It is ready. I will add a test function.

hagenw:
The test function seems to work ok under Octave, but I had to disable the legend as it covered the whole figure. I guess you have split it to six different subplots in order to not have to many overlapping points, or are the subplots showing different things?
For me they looked very similar, here is the first one of them, I guess this is the desired result?
sphbessel_zeros

hagenw:
I was looking to update the reference as the Campos and Calderón paper was published in 2012, see http://doi.org/10.1155/2012/873078
The problem is that they reformulated some of their equations and I was not able to fit them easily to the ones in https://arxiv.org/pdf/1105.0957.pdf, which should be the paper you referenced. Could you have a look at it and propose if we should stay with the arxiv version or switch to the other one?
I would then do the update to the comments.

I'm also wondering if we should add a link to the [scipy routine](https://github.com/scipy/scipy/blob/v0.19.1/scipy/signal/filter_design.py#L3666-L3702) as you used it as a template.

Otherwise, I'm really happy that this seems to be finally solved as it bothered me since the first version that included NFC-HOA.

fietew:
> I was looking to update the reference as the Campos and Calderón paper was published in 2012, see http://doi.org/10.1155/2012/873078
The problem is that they reformulated some of their equations and I was not able to fit them easily to the ones in https://arxiv.org/pdf/1105.0957.pdf, which should be the paper you referenced. Could you have a look at it and propose if we should stay with the arxiv version or switch to the other one?
I would then do the update to the comments.

Yes, I also had a look at the journal paper, which covers a more generalised problem and is harder to understand. I would stay with arxiv version since the formulas in the ``campos_zeros`` correspond directly to equations in the paper.

> I'm also wondering if we should add a link to the scipy routine as you used it as a template.

Yes, that's a good idea.

hagenw:
Ok, I updated the references accordingly.

For the test function, could you maybe send me a screen shot of the Matlab result, that I can try to get the same with Octave.

SFS_general/sphbesselh_zeros.m
```diff
@@ -53,74 +66,114 @@
...
+    return
+end
+
+% Approximate roots of nth-order ordinary Bessel polynomial as starting points
+r = campos_zeros(order);
+
+% Zeros of nth-order ordinary Bessel polynomial are the same as for
+% exp(1/x)*K_{n+0.5}(1/x), where K_v is the modified Bessel function of second
+% kind. We, hence, define the target function and its first derivative as
+f = @(x) besselk(order+0.5, 1./x, 1);  % modified Bessel function scaled by exp
```
hagenw: I would remove the comment `% modified Bessel function scaled by exp` as it my lead a user to expect the `exp()` function to be part of that line. In addition, I think the comment provided before is sufficient.

fietew: It took me a while, to figure out that ``besselk(nu, x, 1)`` is equal to ``besselk(nu,x)*exp(x)``. This is why I added this comment. But if it misleading, we can remove it.

hagenw: I was a little bit confused by it.
Maybe it helps if we state explicitly the equation in the first comment. Something like
```Matlab
% Zeros of nth-order ordinary Bessel polynomial ... are the same as for modified
% Bessel function of second kind K_v:
% exp(1/x) * K_{n+0.5}(1/x) = ...
% Hence, we define the target function and its first derivative as
```
Where you have to fill the `...` part. Is it also second kind and then `Y`?

hagenw:
I added a numerical test case as well for the order 1, 10, and 100 comparing to the results from the python code. This is not perfect as the outcome depends on the ordering of the zeros in the resulting vector, which can differ for different calculation methods, but I thought it is still better than nothing.

SFS_general/sphbesselh_zeros.m
```diff
@@ -53,74 +66,114 @@
...
-% The following code was used to calculate the zeros with the Multiprecission
-% Toolbox. The results are stored at
-% https://github.com/sfstoolbox/data/tree/master/sphbesselh_zeros
-B = mp(zeros(1,order+2));
-A = B;
-for n=mp(0:order)
-    B(n+1) = factorial(2*order-n)/(factorial(order-n)*factorial(n)*2^(order-n));
+
+function x = newton(f,fp,x0,TOL,MAXITER)
+    % Newton-Rapheson method for approximation of a single root
```
hagenw: Where comes this implementation from?
When I'm looking at [scipy's newton implementation](https://github.com/scipy/scipy/blob/07dd175d5897ebc4dcf8e7f42eb4f2c34f005fe3/scipy/optimize/zeros.py#L65-L184) it looks a lot more complicated.

hagenw:
I also did a NFC-HOA calculation in order to compare with Hahn and Spors (2017) Fig. 8:
```Matlab
conf = SFS_config;
conf.nfchoa.order = 30;
sound_field_imp_nfchoa([-2 2],[-2 2],0,[1.5 1.5 0],'ps',1/conf.c,conf)
```
nfchoa_n30
```Matlab
conf.nfchoa.order = 120;
sound_field_imp_nfchoa([-2 2],[-2 2],0,[1.5 1.5 0],'ps',1/conf.c,conf)
```
nfchoa_n120

It looks very similar, but we have the different of that prominent circular contribution which is not part of the plot in the paper. If I remember correctly we had that discussion already somewhere else, hadn't we?
Is this due to the inclusion/exclusion of the lowest order?

SFS_general/sphbesselh_zeros.m
fietew: The scipy implementation does handle many special cases and has some extensions. Here, the most simple version of the Newton-Rapheson method is implemented.

fietew:
> For the test function, could you maybe send me a screen shot of the Matlab result, that I can try to get the same with Octave.

I am pretty sure, that the coloring is not relevant, but I will explicitly define the colors in the script and send you a screen shot.

> Is this due to the inclusion/exclusion of the lowest order?
What are you referring to? What is the lowest order? What is included/excluded?

fietew:
The plot under MATLAB R2015a:
zeros

SFS_general/sphbesselh_zeros.m
hagenw: Now, I found also in the scipy version. Then maybe the easiest thing is to add not a link to the function header, but simply go with the current help text for `newton()`.

SFS_general/sphbesselh_zeros.m
hagenw: Could you help to fill the `...` please.

SFS_general/sphbesselh_zeros.m
fietew: I can't follow you. What do you want me to do?

hagenw:
The colors in the plot work now also under Octave, but the legend was still far to big. I added a commit that replaces the legend with an updated title, indicating the shown orders for every plot.

hagenw:
Coming back to the computed figure and the comparison to Hahn and Spors (2017) Fig. 8. Forget what I have set about the order. But do you have an idea why we have the obvious difference between those plots (of course besides the difference in amplitude)?

SFS_general/sphbesselh_zeros.m
hagenw: Nothing :)
Just a confirmation that it is fine to not add a reference link to the scipy code for the Newton method.

fietew:
Are you using normalisation for the plot? Also the range of the colormap in  Hahn and Spors (2017) Fig. 8. is [-2,2].

hagenw:
I'm using normalisation, otherwise there is no energy in the plot. But you can do the following to get a very similar energy as in Hahn and Spors (2017) Fig. 8:
```Matlab
conf = SFS_config;
conf.nfchoa.order = 30;
conf.plot.caxis = [-0.3 0.3];
sound_field_imp_nfchoa([-2 2],[-2 2],0,[1.5 1.5 0],'ps',1/conf.c,conf)
```
will result in
nfchoa

This looks similar, but still we have a more pronounced visible circle as indicated by the arrow in the figure.

fietew:
Was #56 the discussion you meant?

fietew:
Is it possible, that you have chosen a different number of loudspeakers? In the paper, 60 has been chosen.
For 64 loudspeaker, one loudspeaker is exactly at 45 degrees, which looks like the virtual point source direction. For 60 loudspeakers, 45 degrees lies between two loudspeakers. Maybe this is the reason.

hagenw:
Ah, brilliant, that's it:
```Matlab
conf = SFS_config;
conf.nfchoa.order = 30;
conf.secondary_sources.number = 60;
conf.plot.caxis = [-0.3 0.3];
sound_field_imp_nfchoa([-2 2],[-2 2],0,[1.5 1.5 0],'ps',1/conf.c,conf)
```
and we will get the same result as in the paper.
nfchoa2

I would say this is then nearly ready. There is only [this discussion on the Bessel polynomial](https://github.com/sfstoolbox/sfs-matlab/pull/152#pullrequestreview-48585974) left, @narahahn should give his go, and after a small commit clean up we could merge.

SFS_general/sphbesselh_zeros.m
fietew: According to https://en.wikipedia.org/wiki/Bessel_polynomials#Definition_in_terms_of_Bessel_functions
the nth-order ordinary bessel polynomial is denoted by ``y_n``, which is also the spherical Neumann function in our nomenclature. I would suggest:

```
% Zeros of nth-order ordinary Bessel polynomial y_n are the same as for 
% the exponentially scaled modified Bessel function of second kind K_v:
% \sqrt{2pi/x} * exp(1/x) * K_{n+0.5}(1/x) = y_n
% Hence, we define the target function and its first derivative as
```

hagenw:
I made another test and tried to improve [Fig. 3.14 from my thesis](https://github.com/hagenw/phd-thesis/tree/master/03_sound_field_errors_and_their_perceptual_relevance/fig3_14) and it worked quite well.

Original NFC-HOA part of the figure:
nfchoa_old

New NFC-HOA part of the figure as calculated with this branch:
nfchoa_new

@fietew: From my side, I have no further change requests for this branch and would like to start cleaning the commit history up, if you agree.

fietew:
I agree.

Issue #133, opened 2017-02-27
Avoid `symmetric` parameter in `ifft()`


hagenw:
The `symmetric` option is used currently in [`interpolate_ir()`](https://github.com/sfstoolbox/sfs-matlab/blob/master/SFS_ir/interpolate_ir.m#L126), compare also #130,  and [`modal_weighting()`](https://github.com/sfstoolbox/sfs-matlab/blob/master/SFS_general/modal_weighting.m#L95). Under Octave this option is not available for `ifft()`, which means we have to replace it.

One solution might be to replace it by `real(ifft(spec));` as it is done in [`signal_from_spectrum()`](https://github.com/sfstoolbox/sfs-matlab/blob/master/SFS_general/signal_from_spectrum.m#L91). I guess one problem might be that this will be not as fast as the other option.

@chris-hld could you please test the performance of `interpolate_ir()` with the old `symmetric` option and the `real(ifft())` version.

chris-hld:
For general signals, the performance is surprisingly similar and fast. For the example I tested (4 channels of a few seconds input), I could not reliably tell, which one is faster.
I am still a bit worried about the output though, since it can be quite different.
As long as everything is correct there shouldn't be any imaginary part in the first place. `'symmetric'` restores and corrects the input, while `real()` just discards everything.

chris-hld:
There is a bug somewhere in `interpolate_ir()`!
The remaining imaginary part of ir is around 0.3 -- this is definitely not round-off errors.
I'll have a deeper look at it, any ideas?

chris-hld:
https://github.com/sfstoolbox/sfs-matlab/pull/153

hagenw:
Just for testing, I implemented a version with an extra function that checks for Octave and uses than `real(ifft(...))`. Under Matlab still `ifft(...,'symmetric')` is used: #154

chris-hld:
https://github.com/sfstoolbox/sfs-matlab/pull/155

hagenw:
Solved by #155

PR #153, opened 2017-06-23
Remove x0 from interpolate_ir() call

chris-hld:
Where did this come from?


hagenw:
´x0´ was removed from ´interpolate_ir()` with #137.
I guess we forgot to remove it from the test function as well?

Do you think that is related to the problem you described in #133?
Could you provide an example for that?

chris-hld:
Only related because the code is not working as is.
The problem of https://github.com/sfstoolbox/sfs-matlab/issues/133 is related to the signal.

PR #154, opened 2017-06-26
Add symmetric_ifft() to fix Octave bug

hagenw:
This is a proposal to solve the Octave bug #133.

This is not really a proper solution, but a workaround as it allows Octave for running through without an error by replacing `ifft(...,'symmetric')` by `real(ifft(...))`.
I don't know yet if it would be desirable to implement the exact same behavior of the Matlab version as `real()` simple disregards the output and is not checking for conjugate symmetric input.

### DIFF
--- a/SFS_general/modal_weighting.m
+++ b/SFS_general/modal_weighting.m
@@ -114,7 +114,7 @@ end
 
 % Inverse DTFT
 if nargout>1
-    Win = ifft([win,zeros(1,order)],ndtft,'symmetric');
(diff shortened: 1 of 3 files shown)


chris-hld:
I would suggest to remove this flag in the future.
As you said, this is more a workaround than a solution.
In general we 'only' have to ensure that the input spectrum really is symmetric.

chris-hld:
https://github.com/sfstoolbox/sfs-matlab/pull/155

hagenw:
This is obsolete and was solved by #155

PR #155, opened 2017-06-26
Downsample full spectrum and use full ifft

chris-hld:
By using a full (symmetric) spectrum, the `symmetric` flag can be avoided.
`real()` only deals with very small round-off errors now.


hagenw:
As we are now using signals that are doubled in length will this slow the calculation down?
For `interpolate_ir()` this might be important as you could imagine calling this function quite often.

By using this approach to avoid `'symmetric'` altogether we also have to change [this line in `modal_weighting()`](https://github.com/sfstoolbox/sfs-matlab/blob/4e87dd3f4fef31e3a7ca08154b8c29e10f91580a/SFS_general/modal_weighting.m#L117).

chris-hld:
The symmetric option in combination with the number of output bins should do exactly the same.
As long as this is not optimized internally, this should make no distance.
There is no indication that this is the case, have a look at the computation times in https://github.com/sfstoolbox/sfs-matlab/issues/133

chris-hld:
Should be patched.

hagenw:
Cool, thanks.
I will make a final test next week under Matlab and Octave.

SFS_general/modal_weighting.m
```diff
@@ -114,7 +114,12 @@
 
 % Inverse DTFT
 if nargout>1
-    Win = ifft([win,zeros(1,order)],ndtft,'symmetric');
+    if mod(ndtft,2)
+        win = [win, win(end:-1:2)];
+    else
+        win = [win, win(end-1:-1:2)];
+    end
+    Win = ifft(win);
```
fietew: The new code does not mimic the old code.  try e.g. ``[win, Win] = modal_weighting(10, 1024, conf)`` before and after the changes and check ``Win``

hagenw:
Thanks, `modal_weighting()` works without the `symmetric` option now and returns the exact same results.
I added `test_modal_weighting()` to check for this.

@chris-hld: the new implementation of `interpolate_ir()` does not lead to the same results as before.
Please run `test_interpolation_methods(1)` for the master branch and your branch and compare Fig. 2, 3, 5, and 7. There you can easily find differences.

fietew:
The problem was the phase unwrapping of the second part, i.e. indices greater ``idx_half``, of the spectrum. It is now done for the first part only followed by a mirroring of magnitude and phase.

hagenw:
Cool, now there is now difference any longer.
I would like to squash the commits into two (one for the added `test_modal_weighting()` and one for the changes to `interpolate_ir()`), before merging. Is this fine with you?

fietew:
Yep

chris-hld:
Sure!

On Jul 12, 2017, at 9:14 AM, Fiete Winter <notifications@github.com<mailto:notifications@github.com>> wrote:


Yep

-

PR #158, opened 2017-07-15
Fix checking for correct sample length

hagenw:
In `ir_correct_distance()` we are checking if `conf.N` is long enough for the needed delay.
Delay is in seconds, but it is directly substracted from `conf.N` which is in samples.

This pull request should fix that.

### DIFF
--- a/SFS_ir/ir_correct_distance.m
+++ b/SFS_ir/ir_correct_distance.m
@@ -55,7 +55,8 @@ function ir = ir_correct_distance(ir,ir_distance,r,conf)
 %% ===== Configuration ==================================================
 c = conf.c;
 N = conf.N;
-hrirpredelay = conf.ir.hrirpredelay / conf.fs;
+fs = conf.fs;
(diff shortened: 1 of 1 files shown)


fietew:
It looks good, so far. I would suggest to change the if-condition into ``ir_origlength+delay*fs>conf.N``. It took me a while to figure out the meaning of the current if-condition is. This would be also in better alignment with the ``warning`` function-call afterwards, there it is also ``ir_origlength+delay*fs``.

SFS_ir/ir_correct_distance.m
```diff
@@ -68,10 +69,10 @@
 % Time delay of the source (at the listener position)
 delay = r/c - hrirpredelay; % / s
 % Check if impulse responses are long enough compared to intended delay
-if conf.N-delay<ir_origlength
+if conf.N-delay*fs<ir_origlength
```
fietew: see #158

hagenw:
Thanks, I updated the commit accordingly.

PR #159, opened 2017-07-27
Add elevation to head orientation for ir_* functions

hagenw:
All of the `ir_*` functions in the `SFS_binaural_synthesis` folder were still restricted to azimuth only head orientations of the listener. As all other functions are capable of dealing with elevation, I fixed this and added elevation as well. Now it is also more consistent across the toolbox.
The only functions were head orientation is restricted to azimuthal directions are the ones placed under `SFS_ssr` as the SSR can not deal with elevation yet. For those functions I updated the help message, explicitly stating that we deal with azimuthal head orientation.

I updated the documentation for the new head orientations in commit 8faa1cd and added two further commits that fix typos and bugs in the [binaural simulation section of the documentation](http://matlab.sfstoolbox.org/en/ir_head_orientation/binaural-simulations/).


fietew:
Maybe it would be a good idea to support both cases, i.e. 1st only ``phi`` given and 2nd ``[phi theta]`` given? This would ensure backwards compatibility. I think this could be done quiet easily by checking the length of the input.

hagenw:
I think this is already implemented. All the `ir_*` functions will pass the head orientation to `ir_generic()`, this [passes it to `get_ir()`](https://github.com/sfstoolbox/sfs-matlab/blob/c19494f3a47ee9ad3d124028b6ef0bd7448689d5/SFS_binaural_synthesis/ir_generic.m#L84) and that [checks already for azimuthal head orientation only](https://github.com/sfstoolbox/sfs-matlab/blob/ir_head_orientation/SFS_ir/get_ir.m#L88-L90).

fietew:
Ah, okay, cool.

hagenw:
Is this ready for merge or do you want to do more testing?

PR #164, opened 2017-08-15
Update comments and references in SFS_config

hagenw:
I cleaned up a few comments in the configuration and added missing references.

### DIFF
--- a/SFS_config.m
+++ b/SFS_config.m
@@ -1,5 +1,5 @@
 function conf = SFS_config()
-%SFS_CONFIG Configuration file for the SoundFieldSynthesis functions
+%SFS_CONFIG returns a struct with the default configuration settings
 %
 %   Usage: conf = SFS_config
 %
@@ -11,9 +11,9 @@ function conf = SFS_config()
 %   them in your script (e.g. conf.fs = 48000) and pass the conf struct to
 %   the desired function as last input (e.g. tapering_window(x0,conf)).
 %
-%   So edit this function only, if the default values have changed!
+%   Please don't edit this function to change your configuration settings!
 %
-%   see also: SFS_start
+%   See also: SFS_start
 
 %*****************************************************************************
 % The MIT License (MIT)                                                      *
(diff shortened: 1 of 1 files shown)


SFS_config.m
```diff
@@ -76,17 +74,16 @@
...
 conf.tmpdir = '/tmp/sfs'; % string
-% Debugging level. We are supporting 3 levels:
+% Debugging level. We are supporting two levels:
 %   0 - normal mode
 %   1 - debug modus, showing interim results and plots
 conf.debug = 0; % 0 or 1
-% Show a progress bar in the loops (for example sound_field_mono). This can be
-% useful if you are using secondary sources with >1000 loudspeakers
+% Show a progress bar in selected loops (for example sound_field_mono). This can
+% be useful if you are using a high number of secondary sources
```
fietew: ``, if``

SFS_config.m
```diff
@@ -468,17 +473,34 @@
...
-% AES Convention, Paper 7358, 2008. http://bit.ly/ZCvyQ6
+% Ahrens (2012) - "Analytic Methods of Sound Field Synthesis", Springer,
+% https://doi.org/10.1007/978-3-642-25743-8
+%
+% Ahrens, Spors (2010) - "Sound Field Reproduction Using Planar and Linear
+% Arrays of Loudspeakers", Transactions on Audio, Speech, and Language
+% Processing, vol. 18, no. 8, pp. 2038-2050,
+% https://doi.org/10.1109/TASL.2010.2041106
+%
+% Born, Wolf (1999) - "Principles of Optics", Cambridge University Press, 7th
```
fietew: according to http://opac.lbs-rostock.gbv.de/DB=1/XMLPRS=N/PPN?PPN=559382774 , the 7th Ed. is from 2006. Either the edition or the year is wrong?

hagenw: Maybe your version in Rostock is a reprinted version of the 7th edition. But I have to admit that there counting scheme is a little bit awkward. If you look at the [page of the publisher ](http://www.cambridge.org/catalogue/catalogue.asp?isbn=9780521642224), you see that the first release of the 7th edition was 1999, this can also be found in the [extract of the first pages](http://assets.cambridge.org/97805216/42224/sample/9780521642224ws.pdf), which by the way is a reprint from 2003.
I'm not sure what the official way is how to deal with this, but as on the publisher page the use 1999 under classification (left on their page) it is maybe ok to stay with the current entry?

fietew: Yes, stay with the current one.

SFS_config.m
hagenw: Thanks, I fixed that.

PR #169, opened 2017-08-25
add monochromatic version of LWFS-SBL

fietew:
* circular basis expansion of monochromatic plane wave and point source
* extension of ``driving_function_mono_wfs_vss`` to handle plane wave and focused source as virtual secondary sources (similar to time domain implementation)
* add ``driving_function_mono_wfs_pwd`` to handle plane wave decompositions
* add ``pwd_mono_circexp`` to convert circular basis expansion into plae wave decomposition

### DIFF
--- a/SFS_monochromatic/driving_functions_mono/driving_function_mono_localwfs_sbl_ps.m
+++ b/SFS_monochromatic/driving_functions_mono/driving_function_mono_localwfs_sbl_ps.m
@@ -1,19 +1,19 @@
-function [d,delay_offset] = driving_function_imp_localwfs_sbl_ps(x0,xs,conf)
-%DRIVING_FUNCTION_IMP_LOCALWFS_SBL_PS returns the driving signal for a point 
+function D = driving_function_mono_localwfs_sbl_ps(x0,xs,f,conf)
+%DRIVING_FUNCTION_MONO_LOCALWFS_SBL_PS returns the driving signal for a point 
 %source using local WFS with spatial bandwidth limitation
 %
-%   Usage: [d,delay_offset] = driving_function_imp_localwfs_sbl_ps(x0,xs,conf)
+%   Usage: D = driving_function_mono_localwfs_sbl_ps(x0,xs,f,conf)
 %
 %   Input parameters:
 %       x0          - position and direction of the secondary source / m [nx7]
 %       xs          - position of virtual point source / m [1x3]
+%       f           - frequency of the monochromatic source / Hz
 %       conf        - configuration struct (see SFS_config)
 %
 %   Output parameters:
-%       d               - driving signals [mxn]
-%       delay_offset    - additional added delay, so you can correct it
+%       D           - driving function [nx1]
 %
-%   See also: sound_field_imp_localwfs_sbl, driving_function_imp_localwfs_sbl
+%   See also: sound_field_mono_localwfs_sbl, driving_function_mono_localwfs_sbl
 %
 %   References:
 %       F. Winter, N. Hahn, and S. Spors (2017), "Time-Domain Realisation of 
@@ -51,16 +51,14 @@ function [d,delay_offset] = driving_function_imp_localwfs_sbl_ps(x0,xs,conf)
 
 
 %% ===== Checking of input  parameters ========================================
-nargmin = 3;
-nargmax = 3;
+nargmin = 4;
+nargmax = 4;
 narginchk(nargmin,nargmax);
 
 
 %% ===== Configuration ========================================================
 N0 = size(x0,1);
-Nfft = conf.N;
 xref = conf.xref;
-fs = conf.fs;
 % Ambisonics order
 if isempty(conf.localwfs_sbl.order)
     Nce = nfchoa_order(N0,conf);
@@ -75,14 +73,16 @@ else
 end
 % Resolution of plane wave decomposition
 if isempty(conf.localwfs_sbl.Npw)
-    Npw = 2*ceil(2*pi*0.9*fs/conf.c*conf.secondary_sources.size/2);
+    Npw = 2*ceil(2*pi*0.9*f/conf.c*conf.secondary_sources.size/2);
 else
     Npw = conf.localwfs_sbl.Npw;
 end
 
+
 %% ===== Variables ============================================================
-Nlr = ceil(Nce/2)*2;  % order of Linkwitz-Riley Coefficients
-Wlr = fc/fs*2;  % normalised cut-off frequency of Linkwitz-Riley
+Nlr = ceil(Nce/2)*2;  % order of Linkwitz-Riley Filter
+wc = 2*pi*fc;  % angular cutoff frequency of Linkwitz-Riley Filter
+omega = 2*pi*f;  % angular frequency
 
 
 %% ===== Computation ==========================================================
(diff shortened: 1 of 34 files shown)


SFS_monochromatic/driving_functions_mono/driving_function_mono_wfs_vss.m
```diff
@@ -1,20 +1,21 @@
...
 %
-%   Usage: D = driving_function_mono_wfs_vss(x0,xv,Dv,f,conf)
+%   Usage: D = driving_function_mono_wfs_vss(x0,xv,srcv,Dv,f,conf)
 %
 %   Input parameters:
 %       x0          - position, direction, and weights of the real secondary
 %                     sources / m [nx7]
 %       xv          - position, direction, and weights of the virtual secondary
 %                     sources / m [mx7]
+%       srcv        - type of virtual secondary sources
```
hagenw: Could you add the allowed types here like for the explanation of `src` in other driving functions, e.g. `'fs' - focused source`

hagenw:
So far we have always used `1i` as the imaginary number in the toolbox. In your functions you have used `1j`. If you don't mind, you could switch to `1i` as well.

hagenw:
If have tested it so far by running:
```Matlab
conf = SFS_config;
sound_field_mono_localwfs_sbl([-2 2],[-2 2],0,[0 2.5 0],'ps',1000,conf)
sound_field_mono_localwfs_sbl([-2 2],[-2 2],0,[0 1 0],'pw',1000,conf)
```
which resulted in those figures, that look fine to me:
localwfs_sbl_ps
localwfs_sbl_pw

Would it be worth to add a `test_localwfs_sbl()` function parallel to `test_localwfs_vss()`?
In addition, we have `test_exp_imp()`, but nothing for the monaural case.

fietew:
I fixed the point source. Could you please test again?

SFS_monochromatic/driving_functions_mono/driving_function_mono_wfs_vss.m
fietew: fixed in 7c86d9f

hagenw:
The differences are very subtle, but now the figure is exactly symmetric, which was not the case before.
localwfs_sbl_ps2

fietew:
> So far we have always used 1i as the imaginary number in the toolbox. In your functions you have used 1j. If you don't mind, you could switch to 1i as well.

fixed in 7dc3abf

hagenw:
Both testing functions are running, but of course I cannot guarantee that all of the resulting figures look like they should look ;)
For `test_localwfs_sbl(0)` I get a lot of the following warnings, which doesn't happen for `test_localwfs_vss(0)`:
```
Warning: SECONDARY_SOURCE_SELECTION: 0 secondary sources were selected. 
> In secondary_source_selection (line 176)
  In driving_function_mono_wfs_vss>@(X0,XS)secondary_source_selection(X0,XS(1:3),'pw') (line 98)
  In driving_function_mono_wfs_vss (line 108)
  In driving_function_mono_wfs_pwd (line 78)
  In driving_function_mono_localwfs_sbl (line 96)
  In sound_field_mono_localwfs_sbl (line 92)
  In test_localwfs_sbl (line 159) 
```

fietew:
> Both testing functions are running, but of course I cannot guarantee that all of the resulting figures look like they should look ;)
For test_localwfs_sbl(0) I get a lot of the following warnings, which doesn't happen for test_localwfs_vss(0):

The plane wave decomposition used for LWFS-SBL contains angles uniformly distributed on the unit circle. For some (non-enclosing) secondary source distributions, e.g. linear, the propagation direction of some plane waves leads to an empty x0 after secondary source selection.

hagenw:
Should we then disable that warning in the test function or maybe directly in `driving_function_mono_wfs_pwd()`?

fietew:
The results for ``freq_response_localwfs_sbl([0 0 0],[0 2.5 0],'ps',conf)`` do look as expected, but the large magnitude at low frequency is caused by numerical instabilities. For the time-domain implementation of the point source driving function, we used crossover between conventional WFS (low frequencies) and LWFS-SBL (high frequencies) to circumvent this instabilities. Shall this also be implemented for the mono version?

hagenw:
I would say yes. It's always a good ideas if both domains are implemented as similar as possible.

fietew:
What is the purpose of the functions in ``SFS_analysis``? Are they meant to just show a good parametrised example? At the moment, the number of plane waves ``conf.localwfs_sbl.Npw`` is estimated based on the 
given ``f``. It grows linearly with frequency, so it might very high for some iterations in  ``freq_response_localwfs_sbl``. Should we fix the parameters inside ``freq_response_localwfs_sbl``?

hagenw:
If you think there could be better parameter it would be a good idea to fix this in directly `freq_response_localwfs_sbl()`.

For me the functions in `SFS_analysis` have two purposes:
1) They show how you could also calculate impulse and frequency response of your system, without using `dummy_irs()`.
2) The `freq_response_*` functions are a good way to compare the `imp` and `mono` implementations of the toolbox as they calculate the frequency response using the `mono` functions and you can do the same with `dummy_irs()` using the `imp` functions.

hagenw:
For the driving functions it would be a good idea to add a reference.

For `driving_function_imp_localwfs_sbl_ps()` we could add your EUSIPCO paper and for `driving_function_imp_localwfs_sbl_pw()` the one from Nara, correct?
What about the mono-frequent driving functions?

fietew:
> For driving_function_imp_localwfs_sbl_ps() we could add your EUSIPCO paper and for
driving_function_imp_localwfs_sbl_pw() the one from Nara, correct?

No, Nara's paper takes another approach. I will add a reference.

> What about the mono-frequent driving functions?

The monochromatic implementation covers mostly all aspects of the time-domain implementation, so it can be regarded as a simple inverse fourier transform of the latter.

fietew:
Could we do a minor release after merging the PR?

hagenw:
Yes, this shouldn't be a problem. Do you have a few days left as there are a few problems under Octave currently with this PR and I would like to include the Voronoi PR as well.

fietew:
> Do you have a few days left as there are a few problems under Octave currently with this PR and I would like to include the Voronoi PR as well.

I will run the examples at Octave, too. Is there a specific issue with some of the functions? The Voronoi PR will still take a while. We have to cover all the (pathological) cases as done for ``findconvexcone``.

hagenw:
The problems under Octave are related to [this Ocatve bug](http://savannah.gnu.org/bugs/?49996).
The output of `linkwitz_riley()` has not completely symmetric values, which will raise an `could not pair all complex numbers` error in `cplxpair()` which is called by `zp2sos()`. The asymmetry can look like this:
```
-4.64393188087332e+03 - 5.65864736943455e+03i
-4.64393188087333e+03 + 5.65864736943455e+03i
```
where the very last value of the real part of the complex number is not identical.

One workaround would be to add the following at the end of `linkwitz_riley()`
```Matlab
if isoctave
    % Avoid cplxpair error due to numerical problems in Octave,
    % see http://savannah.gnu.org/bugs/?49996
    z = round(1e9*z)*1e-9;
    p = round(1e9*p)*1e-9;
    k = round(1e9*k)*1e-9;
end
```
but I was not able to test if this would still give the same result at the end as under Matlab (as I have no Matlab today).

fietew:
There is an additional octave bug. Try:

```MATLAB
zp2sos([1],[-1; 1], 1)
zp2sos([1; -1],[-1; 1], 1)
```

hagenw:
Your changes seem to have fix the problems.
```Matlab
conf = SFS_config;
freq_response_localwfs_sbl([0 0 0],[0 2.5 0],'ps',conf)
```
Matlab:
matlab
Octave
octave

hagenw:
Regarding the `zpsos()` problem. I haven't looked into the details how you have fixed it, but do you think this `zpsos()` call in [`bilinear_transform()](https://github.com/sfstoolbox/sfs-matlab/blob/master/SFS_general/bilinear_transform.m#L64-L67) is affected as well?

hagenw:
I ran the `test_localwfs_sbl()` under Octave and Matlab and it worked well in both cases. I added another commit to disable the zero secondary source selection warning in the pwd driving functions. If you think that is not a good idea we can revert this.
Otherwise this PR is ready in my opinion and I would clean up the commit history and merge if this is fine with you.

fietew:
FYI: https://savannah.gnu.org/bugs/?51936

> Regarding the zpsos() problem. I haven't looked into the details how you have fixed it, but do you think this zpsos() call in `bilinear_transform() is affected as well?

I think any usage for zp2sos() is affected by this problem, because it seems to occur whenever you have an odd number of zeros.

SFS_monochromatic/driving_functions_mono/driving_function_mono_localwfs_sbl_ps.m
```diff
@@ -107,20 +107,24 @@
...
 % Lowpass filtering
 [sos,g] = zp2sos(zlp,plp,klp,'down','none');  % generate sos
-Hlp = g.*prod( (sos(:,1:3)*[-omega.^2; 1i*omega; 1]) ...
-      ./ (sos(:,4:6)*[-omega.^2; 1i*omega; 1]),1);
+if isoctave
+  % WORKAROUND: there seems to be an Octave bug
+  Hlp = g.*prod( (sos(:,3:-1:1)*omega_vec)./(sos(:,4:6)*omega_vec),1);
+else
+  Hlp = g.*prod( (sos(:,1:3)*omega_vec)./(sos(:,4:6)*omega_vec),1);
+end
```
hagenw: Is this to deal with https://savannah.gnu.org/bugs/?51936?

fietew: Yes, dunno, if there exists a better workaround.

hagenw: I don't have a better one. I will add the link to the bug in the comment there.
Is this otherwise ready for clean up, or are you planing to push more to this PR?

fietew: ready for clean up

hagenw:
I have updated the commits, if you agree with them I will merge it into master.
I would like to put also some entries in the NEWS file. Would the following fit?
```
- add monochromatic implementation of LWFS using spatial bandwidth-limitation
- add monochromatic circular expansion functions for ps and pw
- add function for conversion from circular to plane wave expansion
- add freq_response_* and time_response_* for all LWFS methods
```

fietew:
> I have updated the commits, if you agree with them I will merge it into master.
I would like to put also some entries in the NEWS file. Would the following fit?
> ```
> - add monochromatic implementation of LWFS using spatial bandwidth-limitation
> - add monochromatic circular expansion functions for ps and pw
> - add function for conversion from circular to plane wave expansion
> - add freq_response_* and time_response_* for all LWFS methods
> ```

Yes

PR #176, opened 2017-11-20
DOC: streamline README and add extra usage

hagenw:
This pull request is intended to clean things up for a change on the navigation in the documentation.
The changes I would like to do for the navigation can be tested here: http://matlab.sfstoolbox.org/en/nav/.

There you can easily switch between Theory, Matlab, and Python via the new navigation menu to the top left. You may also realize that by switching to Python or Theory the sidebar menu is all black at the beginning, which is not the case for Matlab, which is also highlighted by the following images.

matlab

python

This pull request proposes code to achieve the same result as for the other two, see http://matlab.sfstoolbox.org/en/sphinx/.
It has one disadvantage: I had to shorten the README and move most of its content to the usage section in the documentation. @fietew is this ok with you or would you prefer to have a long README?


hagenw:
To achieve the clean up of the left navigation bar, we actually don't have to change the README. The same can be achieved by the changes proposed in #177.

You can compare the result of this PR
http://matlab.sfstoolbox.org/en/sphinx/
with the one from #177 
http://matlab.sfstoolbox.org/en/sphinx2/

hagenw:
@fietew: I would like to merge #180 in the near future and it would be nice if you could say your preference on this PR or the alternative implementation in #177. It was the question of which information should be in the README file.

fietew:
I would stay with the longer readme from #177.

hagenw:
OK, thanks.

PR #177, opened 2017-12-04
Add hidden toctree to index

hagenw:
This is an alternative implementation to #176.
We don't need to shorten the README, including a hidden toctree in `index.rst` instead of having an additional `contents.rst` file also solves the problem described in #176.

The resulting documentation can be viewed at http://matlab.sfstoolbox.org/en/sphinx2/.


hagenw:
I will close this as it is already included in #180.

PR #178, opened 2017-12-27
Logarithmic spacing for linear secondary source distributions

fietew:
Try:

```MATLAB
close all
clear variables
SFS_start;
conf = SFS_config;

conf.secondary_sources.geometry = 'linear';
conf.secondary_sources.grid = 'logarithmic';
conf.secondary_sources.logspread = 5;

idx = 0;
fdx = 1;
markers = {'diamond', 'square'};
for N0 = [20, 21]
  conf.secondary_sources.number = N0;
  y0 = zeros(N0,1); 
  for spread = [0.1, 0.25, 0.5, 1.0, 1.5, 2.0, 5.0]
     conf.secondary_sources.logspread = spread;
     
     y0(:) = idx;
     figure(1);
     hold on;
     plot(x0(:,1),y0,markers{fdx});
     hold off;
     figure(2);
     hold on;
     plot(x0(2:end-1,1), ...
         (x0(1:end-2,1)-x0(2:end-1,1))./(x0(2:end-1,1)-x0(3:end,1)),[markers{fdx} '-']);
     hold off;
     
     cdx = floor(N0/2)+1;
     fprintf('spread: %2.2f\n', ...
         (x0(end,1)-x0(end-1,1))./(x0(cdx,1)-x0(cdx-1,1)) );
     idx = idx + 0.1;
  end
  fdx = fdx + 1;
end
hold off;
figure(1), title('loudspeaker positions');
figure(2), title('ratio of adjacent loudspeaker spacings');
```

I am still not sure, how to handle the new functionality in the ``conf`` struct: In principle, both logarithmic and equi-distant sampling could be handled using ``conf.secondary_source.logspread``, since ``1.0`` would results in an equi-distant spacing. Should we nevertheless use ``conf.secondary_sources.grid`` in addition?

### DIFF
--- a/SFS_general/secondary_source_positions.m
+++ b/SFS_general/secondary_source_positions.m
@@ -104,35 +104,45 @@ if strcmp('line',geometry) || strcmp('linear',geometry)
     x0(:,2) = X0(2);
     x0(:,3) = X0(3);
     x0(:,5) = -1;  % Direction of secondary sources points in -y direction
-
-    u = linspace(-nls/2,nls/2,nls);
-    du = nls/(nls-1);
-    if isempty(conf.secondary_sources.grid) || ... 
-        strcmp(conf.secondary_sources.grid, 'equally_spaced_points')
-        %          L
-        % x0(u) = --- u + X0
-        %         nls
-        x0(:,1) = X0(1) + u.*L/nls;
-       	% Weight each secondary source by the inter-loudspeaker distance
-        x0(:,7) = L./nls*du;  % dx = x0'(u)*du
+    
+    if isempty(conf.secondary_sources.grid) ...
+            || strcmp(conf.secondary_sources.grid, 'equally_spaced_points') ...
+            || ~isfield(conf.secondary_sources, 'logspread') ...
+            || isempty(conf.secondary_sources.logspread) ...
+            || conf.secondary_sources.logspread == 1.0
+        % equi-distantant sampling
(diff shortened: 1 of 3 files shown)


SFS_general/secondary_source_positions.m
```diff
@@ -100,14 +100,50 @@
...
+        %       L              exp(mu*|n|) + C
+        % x0 = --- * sgn(n) * ----------------- + X0
+        %       2               exp(mu*N) + C
+        %
+        % odd number of loudspeakers:
+        %   n = -N,...,N and N = (nls-1)/2.
+        %   C = -1
+        % even number of loudspeakers:
+        %   n = -N,...,-1,1,...,N and N = nls/2.
+        %   C = -0.5*(exp(mu)+1))
```
hagenw: There is one ")" bracket too many.

hagenw:
Nice addition, could you add two logarithmic example to `test_secondary_source_positions()`, one with `conf.secondary_sources.logspread` < 1 and one with > 1. This would be cool.

Regarding the usage of `conf.secondary_sources.grid`, I think that it should be sufficient to use only `conf.secondary_source.logspread` and have a default value of `1`. If you are planing to add more grid spacings for a linear array then it would make sense to use `conf.secondary_sources.grid`, but otherwise it would add the burden of using two config settings to change one thing.

SFS_general/secondary_source_positions.m
fietew: Fixed in e2d668b

PR #186, opened 2019-03-11
Update http to https and new SFS URLs

hagenw:
This updates the URLs to our new pages https://sfs.readthedocs.io, https://sfs-matlab.readthedocs.io, and https://sfs-python.readthedocs.io.

In addition, when possible I updated other URLs from http to https.

NOTE: the http://sfstoolbox.org URLs for the equation references are not updated yet, this will happen in a new pull request where also the links will be updated.

For `SFS_monochromatic/driving_functions_mono/driving_function_mono_wfs_pw.m` I changed the line endings from dos to unix, that is the reason why the whole file is highlighted as changed.


hagenw:
The `http://sfstoolbox.org/#equation-D.nfchoa.ps.3D` links are the once that I meant with URLs for the equation references that I will update in another pull request. If you see only those then it should be fine.

fietew:
Sorry for the confusion.

Issue #79, opened 2016-04-28
Add modal weighting to NFC-HOA monochromatic


hagenw:
#77 added the ability of modal weighting of NFC-HOA orders to the time domain driving function of NFC-HOA.

This cannot be easily added to the current monochromatic implementation of NFC-HOA, but should be able to be added after implementing the new NFC-HOA structure as proposed in #63.

fietew:
Added in #187.

PR #187, opened 2019-03-13
add modal window to monochromatic nfchoa driving functions

fietew:
Follow up of #79. I think the new structure mentioned there is still in the far future. This PR makes the time and frequency domain driving functions more consistent.


hagenw:
Thanks, looks good.

Should we close #79 after merging this?

PR #189, opened 2019-03-25
Proposed method for interpolation of HRTFs in the time domain

VeraE:
This is a proposal for an alternative HRTF interpolation method in the time domain:

The difference in TOAs is estimated by the maximum of the cross-correlation function. The HRIRs are then shifted to have the same TOA, interpolated and then the resulting HRIR is shifted to the interpolated TOA.
The method can handle an arbitrary number of HRIRs as input which is necessary depending on the used method for point selection.
I also added this additional method in `test_interpolation_methods()`. 

What do you think?

### DIFF
--- a/SFS_config.m
+++ b/SFS_config.m
@@ -422,7 +422,7 @@ conf.ir.interpolationpointselection = 'nearestneighbour';
 %                   typical HRIRs with leading and trailing zeros, the error is
 %                   negligible.
 %   'timedomain'  - Interpolation in the time domain with cross-correlation for
-%                   estimation of TOA differences.
+%                   estimation of time of arrival (TOA) differences.
 conf.ir.interpolationmethod = 'simple';
 %
 % If you have HRIRs in the form of the SimpleFreeFieldHRIR SOFA convention, zeros
--- a/SFS_config.m
+++ b/SFS_config.m
@@ -421,6 +421,8 @@ conf.ir.interpolationpointselection = 'nearestneighbour';
 %                   see test_interpolation_methods.m in the validation folder. For
 %                   typical HRIRs with leading and trailing zeros, the error is
 %                   negligible.
+%   'timedomain'  - Interpolation in the time domain with cross-correlation for
+%                   estimation of TOA differences.
 conf.ir.interpolationmethod = 'simple';
 %
 % If you have HRIRs in the form of the SimpleFreeFieldHRIR SOFA convention, zeros
--- a/SFS_ir/interpolate_ir.m
+++ b/SFS_ir/interpolate_ir.m
@@ -146,7 +146,8 @@ if useinterpolation && length(weights)>1
             ir_shifted = zeros(N+max_shift,size(ir,1));
             for nn = 1:size(ir,1)
                 ir_shifted(:,nn) = [zeros(TOA_diff_to_last(nn),1); ...
-                    ir_upsampled(:,nn); zeros(max_shift-TOA_diff_to_last(nn),1)];
+                    ir_upsampled(:,nn); ...
+                    zeros(max_shift-TOA_diff_to_last(nn),1)];
             end
             
             % Interpolate aligned irs
--- a/SFS_ir/interpolate_ir.m
+++ b/SFS_ir/interpolate_ir.m
@@ -26,7 +26,7 @@ function ir = interpolate_ir(ir,weights,conf)
 %     'freqdomain'  - Interpolation in the frequency domain performed separately
 %                     for magnitude and phase.
 %     'timedomain'  - Interpolation in the time domain with cross-correlation for
-%                     estimation of TOA differences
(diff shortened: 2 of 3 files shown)


SFS_ir/interpolate_ir.m
```diff
@@ -25,6 +25,8 @@
 %                     responses.
 %     'freqdomain'  - Interpolation in the frequency domain performed separately
 %                     for magnitude and phase.
+%     'timedomain'  - Interpolation in the time domain with cross-correlation for
+%                     estimation of TOA differences
```
hagenw: For new users it might help if you write `time of arrival (TOA)` here instead of `TOA`

SFS_config.m
```diff
@@ -421,6 +421,8 @@
 %                   see test_interpolation_methods.m in the validation folder. For
 %                   typical HRIRs with leading and trailing zeros, the error is
 %                   negligible.
+%   'timedomain'  - Interpolation in the time domain with cross-correlation for
+%                   estimation of TOA differences.
```
hagenw: For new users it might help if you write `time of arrival (TOA)` here instead of `TOA`

SFS_ir/interpolate_ir.m
```diff
@@ -122,6 +124,43 @@
         ir = ifft(magnitude.*exp(1i*phase),[],3);
         % Avoid round-off errors
         ir = real(ir);
+    case 'timedomain'
+        % See Hartung et al. (1999)
```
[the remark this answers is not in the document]
VeraE: The publication is listed in the comment/help section at the start of the function with a link to the AES library. This should be fine?

SFS_ir/interpolate_ir.m
VeraE: Good point, I changed that.

SFS_config.m
VeraE: Changed that as well, thanks for the comments!

SFS_ir/interpolate_ir.m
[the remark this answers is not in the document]
hagenw: Yes, you are right, I didn#t saw that.

SFS_ir/interpolate_ir.m
```diff
@@ -122,6 +124,43 @@
...
+            k = -(N-1):(N-1);
+            shifts = k(idx_max);
+            
+            % Shift all irs back to align with last ir
+            max_shift = max(shifts)-min(shifts);
+            TOA_diff_to_last = shifts-min(shifts);
+            ir_shifted = zeros(N+max_shift,size(ir,1));
+            for nn = 1:size(ir,1)
+                ir_shifted(:,nn) = [zeros(TOA_diff_to_last(nn),1); ...
+                    ir_upsampled(:,nn); zeros(max_shift-TOA_diff_to_last(nn),1)];
```
hagenw: This line is a little bit too long, maybe you change it to
```matlab
                ir_shifted(:,nn) = [zeros(TOA_diff_to_last(nn),1); ...
                                    ir_upsampled(:,nn); ...
                                    zeros(max_shift-TOA_diff_to_last(nn),1)];
```

SFS_ir/interpolate_ir.m
```diff
@@ -122,6 +124,43 @@
...
+            end
+            
+            % Interpolate aligned irs
+            ir_interp = sum(bsxfun(@times,ir_shifted,weights'),2);
+            
+            % Interpolate TOA differences and shift interpolated ir accordingly
+            TOA_diff_interp = round(TOA_diff_to_last*weights);
+            ir_interp = ...
+                ir_interp(TOA_diff_interp+1:end-(max_shift-TOA_diff_interp));
```
hagenw: This is cropping the interpolated impulse response and thereby shifting further to beginning by `TOA_diff_interp`, correct? Why is this step needed?

SFS_ir/interpolate_ir.m
VeraE: Ok, changed that. Looks clearer now.

SFS_ir/interpolate_ir.m
VeraE: First all irs have been shifted further to the end to align them with the ir that arrives the latest by adding zeros at the beginning. The interpolate aligned ir then has to shifted back towards the beginning according to the interpolated TOA difference. This means samples at the start have to be cropped.
The last samples are cropped to preserve the original length of the ir.

VeraE:
Do you know what the problem is with the last two plots in `test_interpolation_methods()` in Octave? I could not see anything special in the code that is different from the plots before.

SFS_ir/interpolate_ir.m
hagenw: Ah, ok, now I got it.

hagenw:
It was a strange openGL float data exception.
I will create an additional issue for that.

PR #191, opened 2019-04-11
add for loops for plotting in test_interpolation_methods and fix freq…

fietew:
…uency axis in magnitude spectra plots


hagenw:
Works, thank you.

